PR #25, opened 2015-03-18
Add label and description to input definitions.

jmchilton:
This PR adds them to ``cat3-tool.cwl`` - I can pull these out using the reference implementation data structures and it doesn't seem to affect execution - but I feel like these should have be added to some schema or something - it must be more work than that :).

### DIFF
--- a/examples/draft-2/cat3-tool.cwl
+++ b/examples/draft-2/cat3-tool.cwl
@@ -13,6 +13,8 @@
         {
             "id": "#file1",
             "type": "File",
+            "label": "Input File",
+            "description": "The file that will be copied using 'cat'",
             "commandLineBinding": {"position": 1}
         }
     ],


ntijanic:
> it must be more work than that :)

I think that's all there is to it :)

We can reuse the label/description properties for pretty much anything:
 - top-level tool/workflow object
 - input and output ports
 - nested types in input/output ports
 - workflow steps (as e.g. additional description of why a tool is in the workflow)
 - ...

mr-c:
@tetron Should we add the label/description to the documentation now or wait for your draft 2?

tetron:
I've started the draft-2 spec but it is currently just an outline.  The label and description properties just need to be added to the json-ld context as rdfs:label and rdfs:comment.  If you want to start adding them to examples, go ahead.

ntijanic:
I added them already but used dct:description instead. Change if you thing rdfs:comment is more appropriate.

tetron:
Dublin core metadata is probably more appropriate.  There are probably more dct terms we could be using as well.

PR #32, opened 2015-03-31
initial draft of DebianRequirement

mr-c:


tetron:
@mr-c would it be possible to add some code to the reference implementation that demonstrates how this is intended to work?  I am hesitant to add features without a functioning example.

mr-c:
Can do.

mr-c:
@tetron I've added a first pass implementation; am I hooking into the reference implementation correctly?

mr-c:
From this weeks video chat we concluded that there are two modalities in play here:
1) CWL as a guide on how to use a tool (specifying a list of alternative acceptable installation sources)
2) CWL as documentation of a workflow for reproducibility (specifying a single source to satisfy dependencies)

From the experiences of the Galaxy project we know that people will want to run tools and workflows on a variety of systems and possibly install the tools by hand. Galaxy handles this by allowing admins to mark a tool as self-installed. Galaxy does not verify that the tool exists or even meets the version requirement.

Should implementers check the version number? Even using the [Debian version comparision rules]( https://www.debian.org/doc/debian-policy/ch-controlfields.html#s-f-Version) as I suggested doesn't solve the situation when a tool can't or won't report its version number. Versions could also be mined from package management systems. Feels like a lot of scope creep this early on.

I suggest we add the 'get-version-string' property as a mandatory field (to indicate best practice) with a blank command as an allowable choice.

For non-docker users the reference implementation should support several package management systems; specifically PyPI/pip, Debian, RedHat. They should be added in other pull requests.

ntijanic:
How do you all feel about making that a Requirement? So, something like:

```yaml
class: CommandLineTool
requirements:
 - class: MatchesVersionRequirement
   cmd: "mytool --version"
   pattern: "Version: (.*)"
   matches: ">=1.0 <2.0"
```

Meaning, `mytool --version` needs to report `Version: 1.x`.

Actual docker/package descriptions can be placed in the hints section for this use case. It would serve to check if correct tool is available during runtime, and hints can be used either to automate runtime setup or as additional documentation for administrators.

smoe:
Should we not upload a Debian package of your workflow environment to the experimental section of Debian? To me it would be both a bit of outreach and compatibility test - technically and with the community alike.

mr-c:
Hey @smoe, good to see you here :-)

I think khmer is pending in the NEW queue, though I'm about to release v1.4. Would it be faster to upload to experimental?

mr-c:
Deleting in favor of #240

PR #35, opened 2015-04-16
Draft2 guide

ntijanic:


specification/draft-2/README.md
```diff
@@ -0,0 +1,687 @@
...
+    type: File
+    streamable: true
+    binding:
+      glob: output.txt
+baseCmd: grep
+stdout: output.txt
+successCodes: [0, 1]
+```
+
+Above is defined a command line tool labeled "grep" with two required inputs
```
scottfrazer: "Grep" instead of "grep"?

specification/draft-2/README.md
```diff
@@ -0,0 +1,687 @@
...
+(a file and a string) and one file output.
+
+Input objects contain a `binding` section which describes how the arguments
+are passed onto the command line. In this case, `pattern` and `in_file` are
+positional arguments and `pattern` is passed first, followed by `in_file`.
+Identifiers for inputs and outputs must be unique across the document.
```
scottfrazer: What happens if both have `position: 1`?  Should implementers consider this an error?  What if pattern has `position: 1` and out_file has `position: 3`

specification/draft-2/README.md
ntijanic: Yes. Will fix.

specification/draft-2/README.md
```diff
@@ -0,0 +1,687 @@
...
+
+The `binding` section of the output object contains a single property `glob`
+with a glob pattern describing how to recognize file generated by the tool.
```
scottfrazer: "glob" might not be entirely well-defined.  For example, the Wikipedia entry says:

```
Letter[] matches Letter, but not Letter1
```

However, Python's glob package does not support the `[]` syntax.  I wonder if it's worth enumerating out what the glob operations that are supported?  Or point to an existing Glob implementation for reference (Bash?)  Just a thought... probably not a huge deal because people will probably just use `*` for the most part.

specification/draft-2/README.md
ntijanic: Bindings can have same positions (a common case too - default is 0) and the positions do not need to be increments of 1. It simply affects the sort ordering (less means earlier on the command line). In case of same value for position, the ordering should not matter, but they are sorted by input ID.

I'll add a note to the doc.

specification/draft-2/README.md
ntijanic: Good point. I guess the same applies for JS expressions. Should add notes referencing specific spec/implementation.

specification/draft-2/README.md
tetron: The current algorithm for building the command line considers "position" to be a sort key rather than an absolute position.  If two items have the same position, it breaks the tie based on the lexical sort order of the parent keys ("pattern" and "in_file" in this case) in order to get a deterministic ordering.

out_file isn't part of the input, so it doesn't show up on the command line, so the "position" field would have no effect.

specification/draft-2/README.md
```diff
@@ -0,0 +1,687 @@
...
+    binding:
+      position: 2
+outputs:
+  - id: out_file
+    type: File
+    streamable: true
+    binding:
+      glob: output.txt
+baseCmd: grep
+stdout: output.txt
```
tetron: The current draft supports connecting stdout directly to an output parameter:

```yaml
outputs:
 - id: "#output.txt"
   type: File
   streamable: true
stdout: {ref: "#output.txt"}
```

specification/draft-2/README.md
```diff
@@ -0,0 +1,687 @@
...
+
+Input objects contain a `binding` section which describes how the arguments
+are passed onto the command line. In this case, `pattern` and `in_file` are
+positional arguments and `pattern` is passed first, followed by `in_file`.
+Identifiers for inputs and outputs must be unique across the document.
+
+The `binding` section of the output object contains a single property `glob`
+with a glob pattern describing how to recognize file generated by the tool.
+
+Additionally, `baseCmd` and `stdout` properties describe the executable name
```
scottfrazer: So the base command can be a string, like `python -m mymodule`, but do the docs anywhere say how this ought to be interpreted?  For example, what if the baseCmd were: `first_cmd < some_file && second_cmd 'some string'` and it then took two string parameters.  Implementers shouldn't try to break the string by a whitespace delimiter because of the shell-specific pieces.  Also for the two string parameters tacked onto the end of it, by default are they quoted?

specification/draft-2/README.md
ntijanic: Not sure we really need this. It leaves it to the implementation to determine file name, so we either need to spec the rules or leave it unspecified. Just adds complexity for little benefit imo.

specification/draft-2/README.md
```diff
@@ -0,0 +1,687 @@
...
+inputs:
+  in_file:
+    __class: File
+    path: /local/path/to/file.txt
+    size: 42
+  pattern: find ?me
+allocatedResources:
+  cpu: 1
+  mem: 1024
+```
```
tetron: What's the purpose of __class?  You already know in_file is a File from the input schema.

Currently I have input parameters directly at the top level of the document, I don't think nesting it under "inputs" is really necessary.
```yaml
in_file:
  path:  path: /local/path/to/file.txt
pattern: find ?me
allocatedResources:
  cpu: 1
  mem: 1024
```

specification/draft-2/README.md
ntijanic: Bad wording on my part. If multiple `argv`s supplied, they should be listed as array:

```yaml
baseCmd: [python, -m, mytool]
```

There's an example later in the doc, but i'll add it here as well.

specification/draft-2/README.md
```diff
@@ -0,0 +1,687 @@
...
+label: Grep
+description: Filter lines of text file by regular expression.
+inputs:
+  - id: pattern
+    type: string
+    binding:
+      position: 1
+  - id: in_file
+    type: File
+    streamable: true
```
tetron: What are the semantics of "streamable"?  Is it simply a hint that the file will be read sequentially with no seeking?

specification/draft-2/README.md
ntijanic: Three reasons:
1) Edge cases with unions - an input can be a union of File and some record with `path` property, and we give files special treatment (both in bindings and with making them available to tools and path rewrites).
2) argValue expressions can change the type of the input.
3) If type information comes with the record, we can choose to omit the schema in some cases.

Regarding flattening of Job object - we could do that, but we'd need a special character prefix for non-input fields (e.g. `$allocatedResources').

specification/draft-2/README.md
ntijanic: Yes. Or written sequentially.

specification/draft-2/README.md
scottfrazer: I don't think that quite addresses the point though, because what if I have

```yaml
baseCmd: [first_cmd, &&, second_cmd]
```

This can be interpreted two different ways:

```
first_cmd && second_cmd
```
or
```
"first_cmd" "&&" "second_cmd"
```

ntijanic: You are right, it doesn't :)

It's the second one; implementations may choose not to use command line at all and pass the arguments directly to the process.

I'll try to clarify that.

scottfrazer: That's also a point of contention between my DSL and the CWL spec... in my spec, I interpret it the first way because I believe all commands ought to be executed through the shell.  I leave it up to the user to quote things the way they want... like this:

```
command {
  first_cmd && second_cmd
}
```
versus
```
command {
  first_cmd "&&" second_cmd
}
```

I personally like this a lot better because it leaves it up to the user to quote things properly.  It's worth considering this, I think, because it's a lot more flexible.

specification/draft-2/README.md
tetron: There are a few reasons I'd like to have this.
a) It's shorter, and it makes the connection between stdout and the actual output parameter explicit
b) To address your point about determining the file name, you're right, although in my implementation it determines the file name from the parameter name.
c) It provides feature pairity with stdin, which actually a much more important use case because you can't predict the filename in advance and our previous approach to doing this relied on expressions :-(
```yaml
inputs:
  - id: "#inputfile"
    type: File
outputs:
  - id: "#outputfile"
    type: File
stdin: {ref: "#inputfile"}
stdout: {ref: "#outputtfile"}
```

specification/draft-2/README.md
tetron: Strongly disagree about running all commands through the shell, and in fact, the CommandLineTool is engineered specifically to avoid requiring that all commands run through the shell.  It doesn't mean you can't use the shell, you are just required to be explicit about it, e.g. ["/bin/sh", "-c", "first_cmd && second_cmd"] or use a wrapper script.

a) Putting it on users to quote things properly basically means that most scripts will fall over when they see a file with a space in it because getting quoting correct for all cases is very hard.  If the runtime is doing quoting for you, then it needs to have the command line already broken down into a list of tokens, at which point you might as well run the command directly.
b) It exposes potential security vulnerability to shell injection attacks, due to quoting mistakes and bugs such as [shellshock](https://en.wikipedia.org/wiki/Shellshock_%28software_bug%29).  A CWL tool that's a backend to a web service could be running on untrusted inputs.
c) Running the shell adds a little bit of (unnecessary) overhead compared to invoking the tool directly.

specification/draft-2/README.md
```diff
@@ -0,0 +1,687 @@
...
+The tool is described as taking a BAM file and producing an "indexed" BAM file, which is the same file
+but with a .bai file "attached" to it. It also sets the metadata key `cwlm:format` to "BAM".
+
+Metadata keys can be any strings, but we encourage use of URIs, specifically those in the cwlm namespace
+(which is still not defined - work in progress).
+
+The `secondaryFiles` property is a list of file extensions in form of `.ext` that get appended to
+path of master file. This tells the CWL implementation to always make those files available in the
+same directory as the master file. To replace the extension (rather than append) use `^.ext` -
+e.g. `^.dict` for FASTA files.
```
tetron: I started working on the secondaryFile problem last night.  My proposed solution is a bit more general, using regular expressions:

```yaml
    secondaryFiles:
      - match: "(.*ref\.fasta)"
        replace:
          - "\1.amb"
          - "\1.ann"
          - "\1.bwt"
          - "\1.pac"
          - "\1.rbwt"
          - "\1.rpac"
          - "\1.rsa"
      - match: "(.*)ref\.fasta"
        replace:
          - "\1chromsizes.tab"
          - "\1mergedref.bed"
          - "\1ref_map.tab"
```

What do you think?

specification/draft-2/README.md
tetron: Is __class required or just for disambigiouation?  Maybe it should be "@type": "cwl:File" ?

What if allocated resources is handled separately?  job.resources.yml alongside job.input.yml?  The allocated resources are sort of a pseudo-input, in some cases (like debugging a job) you want to know that two jobs are the same except for their resource allocation, in other cases (you just want to know if two jobs had the same input) the resource allocation is irrelevant.

specification/draft-2/README.md
ntijanic: What do you think about an stdin/stdout boolean in bindings to accomplish the same thing?

At least the stdin one. Still unsure about stdout due to the naming issue.

specification/draft-2/README.md
ntijanic: It would be equivalent to `"@type": "cwl:File"`, but avoided the `@` due to Avro.

I think allocatedResources should be accessible in the expressions/wrappers, but no real preference how to encode them. It would also be great to have the ability to add other fields later on.

It doesn't matter much how we encode this, it's just that having a single file/object with all info seems simplest (in all cases implementations should be able to distinguish inputs from additional context like resources). No preference on nested vs flat-with-prefixes.

specification/draft-2/README.md
ntijanic: It's much more general, but is it needed? In my experience, most of these cases are covered with just e.g. ".ref_map.tab" or "^.fa.ext" (append or replace). Did you find examples where it doesn't work?

Nothing stopping us from having both options (depending if array item is string or object).

specification/draft-2/README.md
scottfrazer: Doing `["/bin/sh", "-c", "first_cmd && second_cmd"]` won't work because once you tack on the parameters it falls apart.  For example

```
["/bin/sh", "-c", "ls && ls", "/etc"]
```

That's not the same thing as `["/bin/sh", "-c", "ls && ls /etc"]`

I just feel like it's hard to mandate that people nerf their command line interpretations.  I hear your concerns about injection attacks, but if you're accepting tasks from untrusted sources then really all bets are off anyway... for example, I could have a baseCmd of `['bash', 'my_wrapper.sh']` and it takes one string command which I specify as an arbitrary piece of code which I run inside my wrapper script and then I've subverted any attempt at security by auto-quoting.

ntijanic: Is there a workaround with e.g. `$@`?

I'm not that concerned about injection attacks - the whole point is to allow people to execute arbitrary code. Isolation and some sanitation should likely be done on a layer below. 

The quoting is a real concern, but any "tacked on" arguments can be auto-quoted anyway.

tetron: Scott has a good point that using ["/bin/sh", "-c"] it's really hard to build a command line that uses the shell and also uses input bindings.  So, how about a "shell: true" flag, which permits auto-quoting?

Regarding injection attacks, the situtation I have in mind is one where you have some online service where users are submitting input that runs a pre-defined workflow, where the user doesn't necessarily have access to the code or even all the input data.  For example, analysis of a population where the individual genomes are private but releasing aggregate statistics is consented.

specification/draft-2/README.md
tetron: Yes, a stdin boolean would accomplish the same thing.  Presumably we need to specify that having `stdin: true` on more than one parameter is an error, unless we want to allow concatenated inputs (in which case there's a question of ordering).

For setting the output file, what about having an alternative to glob with exact-match semantics?

```yaml
outputs:
  - id: "#bam"
    type: File
    value: {path: "output.bam"}
    stdout: true
```

specification/draft-2/README.md
tetron: Here is the full parameter definition that is taken from.  Note I haven't implemented `sub` yet so this is still a proposal:

```yaml
  - id: "#reference"
    type: File
    description: "lobSTR's bwa reference files"
    match: ".*ref\.fasta"
    format: "edam:format_1929"
    secondaryFiles:
      - match: "(.*ref\.fasta)"
        replace:
          - "\1.amb"
          - "\1.ann"
          - "\1.bwt"
          - "\1.pac"
          - "\1.rbwt"
          - "\1.rpac"
          - "\1.rsa"
      - match: "(.*)ref\.fasta"
        replace:
          - "\1chromsizes.tab"
          - "\1mergedref.bed"
          - "\1ref_map.tab"
    commandLineBinding:
      prefix: "--index-prefix"
      sub:
        match: "(.*)ref\.fasta"
        replace: "\1"
```

This uses regular expressions in three separate places:
1. To constrain the input file name `match: ".*ref\.fasta"`
2. To rewrite the input value to just the path prefix `match: "(.*)ref\.fasta"` `replace: "\1"` since that's what the tool wants.
3. To identify the secondary files.

Since we've pulled generalized expressions out of the core, I think it would be really useful to incorporate regular expressions as a core capability,

specification/draft-2/README.md
tetron: Should it just be `"class": "File"` then?  Then  `class` would be a reserved keyword.  We could mark up types defined in "schemaDefs" this way as well.

I agree that allocatedResources should be accessible to expressions/wrappers.  There's also a question of accessing values from allocatedResources, e.g. "foo --threads 4" without using expressions.  Maybe  something like this?

```yaml
baseCommand: foo
arguments:
  - valueFrom: {allocatedResources: "cores"}
    prefix: "--threads"

File: specification/draft-2/README.md (hunk 8 above)
Comment (tetron) [inline specification/draft-2/README.md]: Working on this a bit more, how about this:
You can specify substitutions three ways.  For a primary file called "ref.fasta.tar" you could do one of:
* ".gz" means append to get "ref.fasta.tar.gz"
* "^.idx" means "strip the last file extension and append" so you get "ref.fasta.idx"
* You can specify ^ multiple times, so you could also say "^^.idx" to get "ref.idx"
* Finally, you can use regular expressions:
  * { `match: "(.*)(\.fasta)(\.tar)", replace: "\1\3" }` would yield "ref.tar"

Comment (ntijanic) [inline specification/draft-2/README.md]: Exactly what I had in mind :)

Not sure about adding `sub`. It covers some cases (instead of expressions), but not that many and they don't seem to be prevalent. Same can be said for match+replace form of secondary files.

Can we hold off on adding this at the moment and try without it? It seems several groups will be adapting some actual pipelines, so we can wait to see if it's needed.

File: specification/draft-2/README.md (hunk 6 above)
Comment (ntijanic) [inline specification/draft-2/README.md]: `class` is fine if we reserve it, yes (or at least reserve `class: File`).

Regarding replacement for expressions in core spec, I like the idea of a single input Job object (flat or nested) because we could have e.g. jsonpointer expressions in core spec (rather than some custom notation):

```yaml
baseCommand: foo
arguments:
  - valueFrom:
      class: JsonPointer
      script: "#allocatedResources/cpu"  # or e.g. "#$allocatedResources/cpu" if no nesting
    prefix: --threads
```

specification/draft-2/README.md
scottfrazer: That kind of injection attack is certainly something to take into account.  I think at least having the option to not auto-quote (e.g. `shell: true` like you suggested) is a good idea.  However, that does complicate things a little because now the `stdout` section is WAY harder to implement... like if my command (with `shell: true`) is "do_stuff > output.txt", then it's probably not possible for a workflow engine to honor the "stdout" property properly.

ntijanic: Yeah, the whole system was conceptualized for creating argv+redirection for a single process (which could be encoded as a command line). 

Still not sure what the behavior would be, even if stdin/stdout were ignored:
 - Would setting the shell flag disable all quoting?
 - Would there be a per-input flag to enable/disable quoting? If someone wants to construct e.g. a command with pipes and redirection, the "special" arguments shouldn't be quoted but other things (inputs) should.
 - Do we simply ignore stdin/stdout properties if shell is enabled? We could just treat them as appending to the command line.

specification/draft-2/README.md
```diff
@@ -0,0 +1,687 @@
...
+    type: ['null', integer]
+    binding:
+      prefix: --context
+# ...
+```
+
+Above, type of these two inputs is declared as union of "null" (note this must be a string)
+and their actual type to signify that these inputs are optional.
+
+Bindings include the `prefix` property that specifies these should be passed to
```
scottfrazer: Regarding this `prefix`... I just want to make sure, do users have control to append a trailing space to the end of the prefix?  e.g. can I do `prefix: "--context "` as well as `prefix: "--context="`?

If so it might be worth mentioning that

ntijanic: No, we've split that into `prefix` and `separator`, again for going around command line. They can ignore the `separator` and have `prefix: --context=` but not with trailing space. I should make that clear.

Perhaps it's worth changing it to just `prefix` (as in the DSL), but it doesn't play well with UI editors (the trailing space).

specification/draft-2/README.md
tetron: @scottfrazer I'm not sure how stdin and stdout handling is any different?  Unless you are assuming that the script gets piped to the shell.  I'm proposing that the shell command is placed in the shell invocation command line like this:

```/bin/sh -c "first_cmd | second_cmd"```

then it connects the shell's stdin to `first_cmd` and  `second_cmd` to the shell's stdout.  The runtime would continue to handle stdin/stdout redirection around the `sh` process exactly the same way it does now.

@ntijanic I think the only reasonable/safe approach is to quote everything by default, and provide a per-input flag to disable quoting explicitly.  Example:

```yaml
shell: true
arguments:
  - first_cmd
  - "my file"
  - {value: "|", shellquote: false}
  - second_cmd
```

It builds the command line as a list of tokens (existing behavior), quotes each token (unless `shellquote: false`) then joins with spaces to get the literal string that will be passed to the shell.

This would result in an argv something like:

```json
["/bin/sh", "-c", "first_cmd my\\ file | second_cmd"]
```

specification/draft-2/README.md
tetron: So `separator` can be either empty string (no separator), a space (a special case indicating a token boundary) or another token indicating the start of the "value" part of the argument (almost always '=' by convention).  I also feel like there should be a more compact/intuitive way to express that, but I'm not entirely sure what.

specification/draft-2/README.md
tetron: You are right, for the last case it would better to accomplish it through something like a RegexExpressionEngine, and allow "secondaryFiles" to be generated from arbitrary expressions if necessary.

specification/draft-2/README.md
scottfrazer: If I'm understanding you correctly, then if my baseCmd were `ls > myfile`, and I did `shell:true`, then it'd execute as `/bin/bash -c "ls > myfile"`?  If that's the case, then the `stdout` declaration in the task would be pointing to an empty file... which I guess is okay but perhaps confusing / misleading

specification/draft-2/README.md
scottfrazer: I'd have to argue against "almost always '=' by convention".  In our experience it's a very equal mix based on whatever the argv parser for that particular language happens to accept (or a lot of custom hand-written argv parsers).

Also, I'm not quite clear about where separator even exists... All I could find in the docs was:

"Separator to insert between prefix and the actual value. If not specified, null, or a single space " ", then the prefix and value must be added as separate command line arguments; for any other value (including the empty string ""), the prefix, separator and value must be concatenated into a single command line argument."

I just don't really understand why it's not part of the prefix?

specification/draft-2/README.md
tetron: The main advantage of the `{ref: "#input_item"}` notation is that it plays nicely with JSON-LD (because `ref` is an alias for `@id`) by linking the input node to the node where it is used.  That is also the reason I prefer the `stdout: {ref: "#output_item"}` notation.

Unfortunately, it isn't as obvious how to do that with `allocatedResources` because it doesn't have a good placeholder in the graph to link to the way the input and output items do.

specification/draft-2/README.md
tetron: If possible, I think we should try deemphasize globbing outputs in favor of having deterministic outputs (see below).  While there will be cases where predicting the output file names in advance is difficult or impossible (extracting an arbitrary tar file for example), most cases we know exactly what outputs we are expecting and it's better to match then exactly.

specification/draft-2/README.md
tetron: So the thing we're really trying to indicate is whether the prefix and value get concatenated into a single token `["-lfoo"]` or are separate tokens `["-l", "foo"]`.  You probably don't want to write `prefix: "-l "` and get `["-l foo"]` as a single token.

I think maybe instead of the current behavior of "separator" we should just be explicit about this?

```yaml
prefix: -l
value: foo
join: false # (default)
```
results in `["-l", "foo"]`

```yaml
prefix: -l
value: foo
join: true
```

results in `["-lfoo"]`

```yaml
prefix: -l=
value: foo
join: true
```

results in `["-l=foo"]`

specification/draft-2/README.md
tetron: Yes, although in your example you could use `glob: "myfile"` in `outputBinding` to indicate that it it should be captured as an output file.  But it is preferrable to capture it explicitly:

```yaml
stdout: myfile
baseCmd: ["ls"]
```

The reasoning for being explicit (in addition to the desire to avoid dependence on the shell) is that it gives the workflow platform more options for managing `stdin`  and `stdout`, for example streaming to/from a remote storage service instead of staging it on the local disk.

specification/draft-2/README.md
ntijanic: No preference between prefix+separator and prefix+join. Slightly prefer these two to just prefix (with trailing space meaning join=false) since trailing spaces don't play well with visual editors, but this can be worked around.

On the other hand, if we use Scott's notation, we remove a whole property so that's good :)

specification/draft-2/README.md
ntijanic: Not sure I like the port references. It seems okay for simple cases, but cannot handle allocated resources (a very common case) and cannot pick a value from a complex-type input port.

It seems okay to require expressions to pass allocated resources to command line. Not okay for stdin redirection, but a boolean "stdin" flag on input ports seems easier to understand.

specification/draft-2/README.md
ntijanic: > Presumably we need to specify that having stdin: true on more than one parameter is an error

Yeah, and only on `type: File` ports.

> For setting the output file, what about having an alternative to glob with exact-match semantics?

Not sure how useful it is since not all tools will have deterministic output names. Can we hold this off for another version? Feels like we're adding too many features without use cases to support them.

tetron: Okay, let's have `stdin` and `stdout` be a flag.  There will need to be an additional validation step to check that it shows up only zero or one times and only on a File type input or output.

For stdout, either we have the implementation choose the filename, or we need to specify it somehow.  Perhaps by specifying `valueFrom` with a simple string literal?  The user can also use `valueFrom` with an expression to generate more complex filenames.

tetron:
@ntijanic @scottfrazer 

Summary of proposed changes from this discussion:

* `stdin` and `stdout` become flags on File input parameters.
* Use `valueFrom` to specify deterministic output filenames via string literal or expression.  Name from `valueFrom` will be used for stdout.
* `shell` (default false) on command line tools.  If true, this will builds the command line as normal but runs it as a shell command instead of directly `exec()`
* `shellquote` (default true) on commandLineBinding  to indicate whether fragments text that should be quoted or included in the shell command as-is, allowing pass-through of as shell metacharacters such as pipes.
* Simple syntax (described above) for for constructing filenames of secondary files based on the primary file name.  Can also use an expression to generate secondary file names.
* Instead of `separator` which is a character value, have a flag `separate` (default true) which indicates whether the prefix and value are separate argv entries or concatenated into a single entry.

mr-c:
What's the overall status of this guide?

tetron:
I'm not sure what we want to do, at minimum it needs to be brought up to date with draft 2.

PR #48, opened 2015-05-28
fix no method error on unrecognized class

porterjamesj:
Before `makeTool` would just return `None` on an unrecognized class, resulting in a no method error as soon as anyone else tried to use it. This just makes the problem more clear/explicit.

I was somewhat surprised that this wasn't caught by an avro validation. Looking around, I can't find a clear place where the complete tool definition is validated by the CWL avro schema. Is there a function in the library to do this in a general way (e.g. if I just wanted to load a tool definition as a python dict and validate it without actually running it)? If not what would it look like?


tetron:
Thanks, "raise WorkflowException" is now the default behavior of makeTool() if it didn't determine which tool type it has.

Also, validation is now much more rigorous.  Loading a workflow now validates the entire workflow up front, instead of deferring validation of each step until execution.

PR #61, opened 2015-06-12
Fix secondary files bug

porterjamesj:
This fixes a bug in using `secondaryFiles`, `schema` was being used to get the list of secondaryFiles to use rather than `binding`. There probably should be a regression test for this but I don't know how to add one :/ I also noticed an unqualified reference to `ValidationException`, so I fixed that while I was at it.


tetron:
Thanks, I'll incorporate your fix and add tests for `secondaryFiles`

PR #66, opened 2015-06-29
Use pip instead of easy_install

brainstorm:
I had a wacky error otherwise, isn't `easy_install` too bizantine for today's python standards?:

```
$ easy_install
/Users/romanvg/anaconda/lib/python2.7/site-packages/setuptools-14.3-py2.7.egg/pkg_resources/
__init__.py:2512: PEP440Warning: 'llvmlite (0.2.2-1-gbcb15be)' is being parsed as a legacy, 
non PEP 440, version. You may find odd behavior and sort order. In particular it will be 
sorted as less than 0.0. It is recommend to migrate to PEP 440 compatible versions.
error: No urls, filenames, or requirements specified (see --help)
```

### DIFF
--- a/conformance/conformance_test.py
+++ b/conformance/conformance_test.py
@@ -23,6 +23,32 @@ with open(os.path.join(module_dir, args.test)) as f:
 
 failures = 0
 
+def compare(a, b):
+    try:
+        if isinstance(a, dict):
+            if a.get("class") == "File" and not b["path"].endswith("/" + a["path"]):
+                 return False
+            if len(a) != len(b):
+                return False
+            for c in a:
+                if a.get("class") != "File" or c != "path":
+                    if c not in b:
+                        return False
+                    if not compare(a[c], b[c]):
+                        return False
+            return True
+        elif isinstance(a, list):
+            if len(a) != len(b):
+                return False
+            for c in xrange(0, len(a)):
+                if not compare(a[c], b[c]):
+                    return False
+            return True
+        else:
+            return a == b
+    except:
+        return False
+
 def run_test(i, t):
     global failures
     sys.stdout.write("\rTest [%i/%i] " % (i+1, len(tests)))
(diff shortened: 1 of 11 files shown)


brainstorm:
Related to this PR, I tried running `cwl-runner` and got:

```
IOError: [Errno 2] No such file or directory: '/Users/romanvg/anaconda/lib/python2.7/site-packages/cwltool/schemas/draft-2/cwl-avro.yml'
```

Seems that `package_data={'cwltool': ['schemas/draft-1/*', 'schemas/draft-2/*']},` in `reference/setup.py` is not creating nor copying the `.yml` in the right place.

mr-c:
+1

Sounds like a MANIFEST or MANIFEST.in file is needed.

tetron:
@brainstorm Does "python setup.py install" work for you?

The basic problem seems to be that Python disutils has inconsistent handling of symlinks, and `reference/cwltool/schemas` is a symlink to the root `schemas` directory.  I've tested a few variations and here's what I've found:

* `python setup.py install` creates and install a Python egg.  This correctly includes the schema data files.
* `easy_install .` also creates and installs a Python egg.  This correctly includes the schema data files.
* `pip install cwltool` gets a Python source dist (tar.gz).  This also correctly includes the schema data files.
* `python setup.py sdist && pip install dist/cwltool-1.0.20150627015636.tar.gz` also works.
* `pip install .` tries to copy the module files directly into place.  This doesn't dereference symlinks so it fails to install schema data files.

I think the right thing to do is to advise using `python setup.py install` in the documentation.

brainstorm:
Yes! Thanks @tetron and @mr-c, see you in BOSC hopefully!

tetron:
Documentation fixed in abd219fc48b63216dc09f4baa558782b6dfee578, I couldn't merge the pull request directly without it getting messy.

Issue #75, opened 2015-07-08
Enable customizing temp-dir from command line (docker)


robinandeer:
This is an issue on Mac where you can't mount folders (like ``/var``) outside of ``/Users``

brainstorm:
Talking with @tetron and @sinisa88 in the BOSC2015 hackathon, it seems that a reasonable way to fix this is to have a common `tmp` prefix for all the `mkdtmp()` calls so that all steps in the CWL workflow are seen by the boot2docker VM.

guillermo-carrasco:
@robinandeer you can close this :-)

PR #77, opened 2015-07-08
Support for boot2docker 

sinisa88:
Thanks to @robinandeer, @guillermo-carrasco, @brainstorm during BOSC2015 :)

### DIFF
--- a/reference/cwltool/main.py
+++ b/reference/cwltool/main.py
@@ -42,6 +42,9 @@ def arg_parser():
                         default=True, help="Do not delete Docker container used by jobs after they exit",
                         dest="rm_container")
 
+    parser.add_argument("--tmpdir-prefix", type=str, default='tmp',
+                        help="Temp directory prefix for each task or job being executed")
+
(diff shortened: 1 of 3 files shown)


robinandeer:
in response to #75

reference/cwltool/main.py
```diff
@@ -237,20 +240,23 @@ def main(args=None, executor=single_job_executor, makeTool=workflow.defaultMakeT
...
-                       use_container=args.use_container,
-                       pull_image=args.enable_pull,
-                       rm_container=args.rm_container,
-                       rm_tmpdir=args.rm_tmpdir,
-                       makeTool=makeTool,
-                       move_outputs=args.move_outputs)
-        print json.dumps(out, indent=4)
+        pass
     except (validate.ValidationException) as e:
+        print json.dumps(out, indent=4)
```
robinandeer: why print?
this is not future compatible also..

mr-c: Also: the try block appears to be empty now, is that a typo?

PR #78, opened 2015-07-08
add explicit msg when failing during docker req. loading.

hmenager:
probably needs a lot more work...


tetron:
Implemented in 38fe20005fed208c06fe8d8c8799cf70deee5f04

Issue #79, opened 2015-07-09
wc test failing on different platforms


guillermo-carrasco:
Because the output of `wc` is slightly different depending on the platform (linux/mac) or flavour (ubuntu, debian, etc.). The test does fail even if the job executes correctly, because it is checking the expected file size instead of the actual number of words in the file.

mr-c:
@guillermo-carrasco Is this still an issue?

guillermo-carrasco:
Hi @mr-c ,

That was quite a white ago and a lot of things changed. I've re-pulled and re-installed everything, tried to run the tests, and got this:

```
(master) ~/repos_and_code/common-workflow-language (master) ~> ./run_test.sh
readlink: illegal option -- f
usage: readlink [-n] [file ...]
--- Running conformance test draft-2 on  ---
./run_test.sh: line 54: : command not found
Test [1/36] Traceback (most recent call last):
  File "/Users/guillem/anaconda/envs/master/lib/python2.7/runpy.py", line 162, in _run_module_as_main
    "__main__", fname, loader, pkg_name)
  File "/Users/guillem/anaconda/envs/master/lib/python2.7/runpy.py", line 72, in _run_code
    exec code in run_globals
  File "build/bdist.macosx-10.5-x86_64/egg/cwltool/cwltest.py", line 171, in <module>
  File "build/bdist.macosx-10.5-x86_64/egg/cwltool/cwltest.py", line 156, in main
  File "build/bdist.macosx-10.5-x86_64/egg/cwltool/cwltest.py", line 77, in run_test
  File "/Users/guillem/anaconda/envs/master/lib/python2.7/subprocess.py", line 566, in check_output
    process = Popen(stdout=PIPE, *popenargs, **kwargs)
  File "/Users/guillem/anaconda/envs/master/lib/python2.7/subprocess.py", line 710, in __init__
    errread, errwrite)
  File "/Users/guillem/anaconda/envs/master/lib/python2.7/subprocess.py", line 1335, in _execute_child
    raise child_exception
OSError: [Errno 13] Permission denied

1 tool tests failed
```
I have to say though that I haven't spent much time trying to figure out what is the problem. I'm posting it here for the record, in case you know at first sight what may be going on, otherwise I'll try to spot the problem. 

Thanks!

chapmanb:
Guillermo;
[...]

PR #81, opened 2015-07-09
Add support for boot2docker!

guillermo-carrasco:
Thanks a lot to @sinisa88 , @robinandeer and @brainstorm :)


guillermo-carrasco:
We ran all the conformance tests and they all pass... except the wc one, but we opened an issue about it on #79 .

guillermo-carrasco:
Thanks for merging @tetron!

PR #85, opened 2015-07-09
Usability fixes encountered during testing

chapmanb:
- Report a useful error message for docker if we're missing an image and
  don't have a way to retrieve it.
- When running locally, pass environmental variables along to the
  subprocess run. This keeps local PATHs and friends working.
- Avoid removing temporary directories if '--leave-tmpdir' is set.

### DIFF
--- a/reference/cwltool/docker.py
+++ b/reference/cwltool/docker.py
@@ -66,6 +66,9 @@ def get_image(dockerRequirement, pull_image, dry_run=False):
                 if rcode != 0:
                     raise Exception("Docker load returned non-zero exit status %i" % (rcode))
                 found = True
+        elif dockerRequirement.get("dockerImageId"):
+            raise ValueError("Docker image %s not found locally and no mechanism to pull in DockerRequirement: %s"
+                             % (dockerRequirement["dockerImageId"], dockerRequirement))
 
     return found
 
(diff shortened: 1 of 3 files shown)


porterjamesj:
@chapmanb if I understand correctly [it's actually part of the spec *not* to pass environment variables into tool process](https://github.com/common-workflow-language/common-workflow-language/blob/b3da2257a36217078d0a3a5b085928456bce3f86/schemas/draft-2/cwl-avro.yml#L1301-L1302).

chapmanb:
James -- thanks for this. I thought that was only for Docker processes. For local runs, if you scrub the PATH, you'd have highly system specific cwl files pointing to the installation of tools. Maybe there is a way to specify we should pass on the existing value of an environmental variable, but I couldn't identify the way to do this.

tetron:
@chapmanb 

Try providing PATH as an input and use EnvVarRequirement::

```yaml
inputs:
  id: "#PATH"
  type: string
requirements:
  - class: EnvVarRequirement
    envDef:
     - envName: PATH
       envValue:
         engine: "cwl:JsonPointer"
         script: "/job/PATH"
```

ntijanic:
Not sure we should be using `EnvVarRequirement` to bind `PATH` to an input. It might be better to add a `--preserve-environ` switch to the runner for development mode, and specify paths and such in the packaging (currently just in the docker image).

mr-c:
Hey-o, I'm closing this issue as I think we've dealt with all of the mentioned issues.

Thank you @chapmanb for reporting them and for the PR !

PR #89, opened 2015-07-09
Support retrieving docker instances via 'docker import'

chapmanb:
bcbio stores its container as a gzipped download via S3 because it's too big for the docker index. This PR allows pulling in instances similar to docker pull, but via docker import.


mr-c:
+1 for the idea but someone with more docker experience than I have should review the code

PR #97, opened 2015-07-21
Use boot2docker's docker UID if boot2docker is present

### LINKED ISSUE #96: Tools w/ docker req cannot write to output dir on boot2docker (Mac OS X)
I'm trying to build some simple tools that have Docker requirements. I use boot2docker on Mac OS X 10.10.4, and jobs fail because the user inside the container cannot write to the output directory.

This is related to a quirk of boot2docker/VirtualBox folder sharing on Mac OS X. In order to make Mac OS X host directories available as volumes in Docker, [boot2docker uses VirtualBox Shared Folders on Mac OS X](https://github.com/boot2docker/boot2docker/tree/v1.7.1#virtualbox-guest-additions). The /Users directory is shared into the boot2docker VM, and owned by an arbitrary UID:GID (1000:50) inside the boot2docker VM.

I appreciate that cwl-runner tries to prevent root-owned files by using my native UID (501) as the UID inside the container. But this seems to be incompatible with boot2docker shared folders, since the directory in the boot2docker is owned by UID 1000 and UID 501 cannot write to it.

Ironically, letting container processes run as root typically works better on boot2docker/VirtualBox (not that I'd recommend it as a solution), because the container can still write to any directory in the boot2docker VM, and VirtualBox restores ownership back to the host user (501 in my case). 

```
$ id -u
501
$ ls -ldn /Users/dcl9/
drwxr-xr-x+ 98 501  20  3332 Jul 21 11:34 /Users/dcl9/
$ docker run -v /Users/dcl9:/Users/dcl9 ubuntu ls -ldn /Users/dcl9
drwxr-xr-x 1 1000 50 3332 Jul 21 15:34 /Users/dcl9
```

Samples: [gist](https://gist.github.com/dleehr/a32b7449de96e717f9b6)

```
$ cwl-runner copy.cwl copy.json 
/Users/dcl9/Code/envs/cwl/bin/cwl-runner 1.0.20150715173345
[job 4461496720] exec docker run -i --volume=/Users/dcl9/Code/python/cwl-copy/sourcefile:/tmp/job809623956_cwl-copy/sourcefile:ro --volume=/Users/dcl9/Code/python/cwl-copy:/tmp/job_output:rw --volume=/var/folders/7t/q8bb2np92p5b26pk_h5lr0lh0000gn/T/tmpoTa6Hp:/tmp/job_tmp:rw --workdir=/tmp/job_output --user=501 --rm --env=TMPDIR=/tmp/job_tmp ubuntu:latest cp /tmp/job809623956_cwl-copy/sourcefile mycopy
cp: cannot create regular file 'mycopy': Permission denied
Exception while running job
Traceback (most recent call last):
  File "build/bdist.macosx-10.10-intel/egg/cwltool/job.py", line 157, in run
    outputs = self.collect_outputs(self.outdir)
  File "build/bdist.macosx-10.10-intel/egg/cwltool/draft2tool.py", line 401, in collect_output_ports
    ret[fragment] = self.collect_output(port, builder, outdir)
  File "build/bdist.macosx-10.10-intel/egg/cwltool/draft2tool.py", line 437, in collect_output
    raise WorkflowException("No matches for output file with glob: {}.".format(binding["glob"]))
WorkflowException: No matches for output file with glob: {'engine': 'cwl:JsonPointer', 'script': 'job/dest'}.
[job 4461496720] completed permanentFail
Final process status is permanentFail
{}
```

Running the same with the `requirements` removed works as expected:

```
$ cwl-runner copy.cwl copy.json 
/Users/dcl9/Code/envs/cwl/bin/cwl-runner 1.0.20150715173345
[job 4366240144] exec cp /Users/dcl9/Code/python/cwl-copy/sourcefile mycopy
Final process status is success
{
    "output": {
        "path": "/Users/dcl9/Code/python/cwl-copy/mycopy", 
        "size": 13, 
        "class": "File", 
        "checksum": "sha1$cd50d19784897085a8d0e3e413f8612b097c03f1"
    }
}
```

dleehr:
Open to suggestions. This is imperfect because it's possible boot2docker may be installed but not to be used

Fixes #96

### DIFF
--- a/reference/cwltool/job.py
+++ b/reference/cwltool/job.py
@@ -69,9 +69,14 @@ class CommandLineJob(object):
             euid = os.geteuid()
             try:
                 # If running under boot2docker, use the user's uid inside the vm
-                docker_uid = subprocess.check_output(["boot2docker","ssh", "id", "-u"]).strip()
-                if len(docker_uid) > 0: euid = docker_uid
-            except:
+                docker_uid = subprocess.check_output(["boot2docker","ssh", "id", "-u"])
+                if len(docker_uid) > 0:
+                    euid = str(int(docker_uid))
+            except (OSError, subprocess.CalledProcessError, TypeError, ValueError):
+                # OSError is raised if boot2docker doesn't exist
+                # CalledProcessError is raised if boot2docker returns nonzero
+                # TypeError is raised if docker_uid has no len()
+                # ValueError is raised if str/int conversion fails
                 pass
 
             runtime.append("--user=%s" % (euid))
--- a/reference/cwltool/job.py
+++ b/reference/cwltool/job.py
@@ -66,7 +66,15 @@ class CommandLineJob(object):
             runtime.append("--volume=%s:%s:rw" % (os.path.abspath(self.outdir), "/tmp/job_output"))
             runtime.append("--volume=%s:%s:rw" % (os.path.abspath(self.tmpdir), "/tmp/job_tmp"))
             runtime.append("--workdir=%s" % ("/tmp/job_output"))
-            runtime.append("--user=%s" % (os.geteuid()))
+            euid = os.geteuid()
+            try:
+                # If running under boot2docker, use the user's uid inside the vm
+                docker_uid = subprocess.check_output(["boot2docker","ssh", "id", "-u"]).strip()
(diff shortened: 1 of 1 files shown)


dleehr:
I'm refactoring this to support docker-machine too, and will propose a new PR

Issue #112, opened 2015-08-17
Add requirements.txt and .python-version to root


slnovak:
Hi,

It'd be nice to have the following in the CWL codebase:

* A `requirements.txt` file at project root to list Python dependencies needed to run test-suite.

* A `.python-version` to the root to specify the preferred Python version for development. (see [pyenv](https://github.com/yyuu/pyenv).

mr-c:
Moved to https://github.com/common-workflow-language/cwltool/issues/5

PR #113, opened 2015-08-17
add .python-version and requirements.txt

slnovak:
Adds `.python-version` and `requirements.txt` to facilitate development and testing.


mr-c:
@slnovak Thanks. Can you enlightened me as to what tools or standard a `.python-version` file corresponds to? Is this meant to be a minimum required version or something else?

mr-c:
Ah, this is for #112

tetron:
Hi @slnovak I came across this blog post about how to avoid repeating dependencies between setup.py and requirements.txt, what do you think?

https://caremad.io/2013/07/setup-vs-requirement/

slnovak:
Good find!

There's a [good reference](http://stackoverflow.com/questions/14399534/how-can-i-reference-requirements-txt-for-the-install-requires-kwarg-in-setuptool) on SO which shows how you can use PIP's `parse_requirements` to parse through `requirements.txt` for importing requirements into `setup.py`. This might be a better approach for reducing duplicate dependencies.

What do you think about that?

mr-c:
Since the reference implementation is now its own repository I'm going to close this pull request. I agree that is is a good idea! Tracking issue for the new repository is at https://github.com/common-workflow-language/cwltool/issues/5

PR #114, opened 2015-08-26
cwl-avro.yml - added deprectation to terminology

smoe:
Felt that if something is prepared to get in (**may**) then something shall also be prepared to get out again - the deprecation. 

Sidenote: I had felt that this paragraph mixes the properties of a schema too much with properties of an instance of that schema. Please indicate if you want me to patch that.


tetron:
Thanks, this is in the terms section now.

Issue #122, opened 2015-09-08
Workflow passing default value for type record (might be connected with issue #121)


portah:
Tool description:
https://github.com/SciDAP/workflows/blob/master/tools/bedtools-genomecov.cwl
Workflow description:
https://github.com/SciDAP/workflows/blob/master/workflows/scidap/bam-genomecov-bigwig.cwl
Workflow Job:
https://github.com/SciDAP/workflows/blob/master/workflows/scidap/bam-genomecov-bigwig-job.json

Can't run the workflow with error:

```
/usr/local/bin/cwl-runner 1.0.20150814191928
Traceback (most recent call last):
  File "/usr/local/bin/cwl-runner", line 9, in <module>
    load_entry_point('cwl-runner==1.0', 'console_scripts', 'cwl-runner')()
  File "build/bdist.linux-x86_64/egg/cwltool/main.py", line 314, in main
  File "build/bdist.linux-x86_64/egg/cwltool/workflow.py", line 30, in defaultMakeTool
  File "build/bdist.linux-x86_64/egg/cwltool/workflow.py", line 295, in __init__
  File "build/bdist.linux-x86_64/egg/cwltool/workflow.py", line 341, in __init__
  File "build/bdist.linux-x86_64/egg/cwltool/process.py", line 79, in __init__
  File "build/bdist.linux-x86_64/egg/avro/schema.py", line 740, in make_avsc_object
  File "build/bdist.linux-x86_64/egg/avro/schema.py", line 661, in __init__
  File "build/bdist.linux-x86_64/egg/avro/schema.py", line 627, in make_field_objects
  File "build/bdist.linux-x86_64/egg/avro/schema.py", line 342, in __init__
avro.schema.SchemaParseException: Type property "['null', 'null', '#depts']" not a valid Avro schema: null type already in Union
```

tetron:
Should be fixed by #123

PR #123, opened 2015-09-09
null duplication #122

portah:


tetron:
This is not quite right, it should be:

```
if "default" in c and (isinstance(c["type"], basestring) or (isinstance(c["type"], list) and "null" not in c["type"])):
```

PR #156, opened 2015-11-02
Docker socket

tetron:
Add "DockerSocketRequirement" to specification, with test.


tetron:
This feature was rejected for draft-3

PR #157, opened 2015-11-02
File formats

tetron:
Add "format" to File type and FormatOntologyRequirement to spec.


ntijanic:
In cases where a tool can produce/consume multiple formats (e.g. works with BAM or SAM on input and a parameter determines output format), is there a distinction between "produces/consumes formats" annotations on outputs/inputs and "set file format" property on output binding?

It seems the former can be used for (non-fatal?) validation, validating workflows before runs, searching registries and so on, while the latter should be used to set the format property of output files. If the output annotation specifies a single format and there's no format property of output binding, perhaps it can be automatically set (value of Output.format applied to File.format).

Also, if the `@context` registered a prefix like `edam`, it should be possible to use it in the annotation properties (Port.format), but should one be able to use it in OutputBinding.format (which can also be an expression, right)? Can one use namespaces in File.format?

mr-c:
Do you need some prose about the IANA Media Types now that they are the preferred way of representing file formats?

Issue #141, opened 2015-10-09
Proposal: ResourceRequirement


tetron:
```
requirements:
 - class: ResourceRequirement
   coresMin:  1
   coresMax: 4
   ramMin 1024
   ramMax: 4096
   tmpdirMin: 1024
   tmpdirMax: 1024
   outdirMin: 1024
   outdirMax: 1024
```

"min" is the minimum amount of a resource that must be reserved to schedule a job.  If "min" cannot be satisfied, the job should not be run.

"max" is the maximum amount of a resource that the job shall be permitted to use.  If a node has sufficient resources, multiple jobs may be scheduled on a single node provided each job's "max" resource requirements are met.  If a job attempts to exceed its "max" resource allocation, an implementation may deny additional resources, which may result in job failure.

If "min" is specified but "max" is not, then "max" == "min"
If "max" is specified by "min" is not, then "min" == "max".

It is an error if max < min.

If neither "min" nor "max" is specified for a resource, an implementation may provide a default.

cores: reserved number of CPU cores
ram: reserved RAM in megabytes (2**20)
tmpdir: reserved filesystem based storage for the designated temporary directory, in megabytes
outdir: reserved filesystem based storage for the designated output directory, in megabytes

satra:
some scattered thoughts:

in many cases these requirements are closely tied to the application at hand. in nipype we dissociate the spec for a tool, from the runtime requirements for a tool in a specific os/hardware environment for a specific application. it seems this is combing things, which means one would have to rewrite a cwl for a given application and runtime environment.

every job may have a set of necessary dependencies/environments under which it operates (the dependency hell). draft 3 refers to environment variables for environment requirement, not a software dependency chain, and this proposal focuses on some hardware requirements. the following are equally valid resource requirements:

- OSes/kernel versions
- libraries/versions
- etc.,.

what about jobs that require:
- multiple nodes on a cluster, say using mpi (version X)
- overall memory across nodes rather than specific memory per node
- io bandwidth

what about jobs that can run under different hardware environments (e.g., opencl jobs)? would there be a way to specify a preferential order of hardware requirements? 

ps. in many cases the above information could be retrofitted to a CWL based on captured provenance.

tetron:
Hi Satra,

[...]

PR #158, opened 2015-11-02
Resource requirement

tetron:
Add "ResourceRequirement" to spec


mr-c:
For #141

tetron:
Rename `$resources` to `runtime` and move `tmpdir` and `outdir` to it.

PR #212, opened 2016-05-18
[WIP] initial description of cwl-runner interface

mr-c:


mr-c:
I'm merging this for now.

jmchilton:
Would you consider this part of the specification? For instance, would an implementation of the CWL have to allow and respect a ``--no-container`` option? I like the increased documentation - but if this represent what is required to comply with the specification - I'm very uneasy about the level of detail and the implementation details exposed here.

tetron:
@jmchilton I think @mr-c was just documenting what it has now, but I agree I would like to pare this down to the minimum set of options that are useful for interoperability (e.g. the conformance tests).

mr-c:
PRs & discussion accepted, yes :-) But as it stands, this is going into 1.0.

tetron:
@jmchilton Personally I'm responsible for the command lines of `cwltool`, `cwltoil` and `arvados-cwl-runner` and they manage to pass conformance tests while having disjoint sets of additional options.

PR #219, opened 2016-05-24
Corrections and test data

denis-yuen:
* fix a bunch of examples where the filenames do not match between provided samples and command-line calls
* provide data for workflows
* fix a reference to JSON when YML is provided
* a typo or two

### DIFF
--- a/draft-3/UserGuide.yml
+++ b/draft-3/UserGuide.yml
@@ -20,7 +20,7 @@
     - |
       ```
 
-      Use a JSON object in a separate file to describe the input of a run:
+      Use a YAML object in a separate file to describe the input of a run:
 
       *echo-job.yml*
       ```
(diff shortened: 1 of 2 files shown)


mr-c:
Thank you @denis-yuen ! Did you mean to update only the draft-4 user guide? I believe we are only building the draft-3 user guide for the website right now.

denis-yuen:
Not really. 
I was attempting to update the source for http://www.commonwl.org/draft-3/UserGuide.html which is draft-3, but I guess I did draft-4 by accident. I think the changes still apply though.

mr-c:
@denis-yuen Should I open a new PR to add these to the draft-3 user guide or can you update this PR?

denis-yuen:
@mr-c  Give that a shot, I think that handles draft-3 as well

mr-c:
Looks great, thanks!

Issue #215, opened 2016-05-19
SecondaryFiles, streamable and format should not be assignable to type record and enum


mayacoda:
The draft-3 specification states that [CommandInputRecordSchema](http://www.commonwl.org/draft-3/CommandLineTool.html#CommandInputRecordSchema), [CommandInputEnumSchema](http://www.commonwl.org/draft-3/CommandLineTool.html#CommandInputEnumSchema), [InputRecordSchema](http://www.commonwl.org/draft-3/CommandLineTool.html#InputRecordSchema) and [InputEnumSchema](http://www.commonwl.org/draft-3/CommandLineTool.html#InputEnumSchema) have the fields `secondaryFiles`, `streamable` and `format`. These fields are only possible for `type: File` or `items: File` which types record and enum are not. They should be removed from the specification.

mr-c:
Fixed in #232

PR #232, opened 2016-06-17
reparent file specific fields from SchemaBase to Parameter

mr-c:
Proposed fix for https://github.com/common-workflow-language/common-workflow-language/issues/215

Looks like moving the `secondaryFiles`, `streamable`, and `format` fields from `cwl:SchemaBase` to `cwl:Parameter` preserves the intended behavior and removes them places they don't make sense: `{Input,Output}{Record,Enum,Array,,}Schema`

The conformance tests still pass and the docs still build with this change.

### DIFF
--- a/draft-4/Process.yml
+++ b/draft-4/Process.yml
@@ -141,7 +141,29 @@ $graph:
 - name: SchemaBase
   type: record
   abstract: true
+
+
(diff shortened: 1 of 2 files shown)


mr-c:
@tetron What do you think?

Issue #217, opened 2016-05-20
Confusion between CWLType and PrimitiveType in CLT documentation


mayacoda:
There are several places where `PrimitiveType` is used instead of `CWLType`. One example is the definition of `items` in [CommandInputArraySchema](http://www.commonwl.org/draft-3/CommandLineTool.html#CommandInputArraySchema), where the items type includes `PrimitiveType` instead of `CWLType`.

Is there any specific reason for this? What is the purpose of `CWLType` if not to include `File` in these cases?

mr-c:
Good catch, `File` is allowed here and the spec is wrong.

Also `string` is redundant as it is a member of `PrimitiveType`.

`string` should be removed and `PrimitiveType` should be replaced with `CWLType`

mr-c:
re: `string` redundancies; @tetron clarifies that these are for URIs or named SchemaDefs

https://gitter.im/common-workflow-language/common-workflow-language?at=5745a9adec10ddbb09da84f6

mr-c:
@mayacoda I have a fix up at #233

you can preview it by checking out that branch and running:

`cwltool  --no-container --outdir=$PWD/common-workflow-language.github.io site/cwlsite.cwl site/cwlsite-job.json` using the latest cwltool

PR #233, opened 2016-06-17
use CWLType more often

mr-c:
Fixes https://github.com/common-workflow-language/common-workflow-language/issues/217


mr-c:
Needed to enable File types as part of Record types

PR #240, opened 2016-06-27
Dependency hint

mr-c:
Redo of https://github.com/common-workflow-language/common-workflow-language/pull/234

Now: list of specifications with optional identifier field; added test case

### DIFF
--- a/draft-4/CommandLineTool.yml
+++ b/draft-4/CommandLineTool.yml
@@ -627,26 +627,43 @@ $graph:
 
 
 - type: record
-  name: LooseDependencyHint
+  name: SoftwareRequirement
   extends: "#ProcessRequirement"
   doc: |
-    If a CommandLineTool is executed by a platform that is not configured to use
-    Docker, this dependency can annotate hints about the software that should
-    be configured in the environment of the defined process.
+    Software that should be configured in the environment of the defined
+    process.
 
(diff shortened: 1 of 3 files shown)


draft-4/CommandLineTool.yml
```diff
@@ -626,6 +626,44 @@ $graph:
...
+        items: SoftwareSpec
+      doc: "The list of software to be configured."
+
+- name: SoftwareSpec
+  type: record
+  fields:
+    - name: name
+      type: string
+      doc: "The name of the software to be configured."
+      jsonldPredicate: "@id"
```
mr-c: I don't know if this is the correct thing here; was the only was to process the spec due to predicate collision. The other option I thought of was to link it to `doap:name` but that didn't resolve the predicate error.

jmchilton:
Thanks a ton @mr-c - this gets an enthusiastic :+1: from me.

draft-4/draft-4/cat1-tool.cwl
```diff
@@ -5,6 +5,8 @@ description: "Print the contents of a file to stdout using 'cat' running in a do
 hints:
   DockerRequirement:
     dockerPull: debian:wheezy
+  SoftwareRequirement:
+    name: cat
```
jmchilton: Doesn't this need to be an array based on the above definition?

PR #244, opened 2016-07-04
Eliminate dirent

tetron:
Adjust design of Directory type.  Eliminate intermediate `Dirent` record in favor of using `basename` field on File and Directory objects.  Fixes site/ generation workflow.

### DIFF
--- a/draft-4/Process.yml
+++ b/draft-4/Process.yml
@@ -119,6 +119,7 @@ $graph:
         `basename`.
 
         Must be evaluated as an expression only when appearing in InitialWorkDirRequirement.
+      jsonldPredicate: "cwl:basename"
     - name: dirname
       type: string?
       doc: |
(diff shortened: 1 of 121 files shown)


tetron:
Also adds anonymous File literals, similar to anonymous Directory literals.  These are File objects which have a "contents" field which are written out to an actual file as needed.

draft-4/Process.yml
```diff
@@ -84,24 +88,34 @@ $graph:
...
+
+        As a special case, if the `path` field is provided but the `location`
+        field is not, an implementation may assign the value of the `path`
+        field to `location`, and remove the `path` field.
       jsonldPredicate:
         "_id": "cwl:path"
         "_type": "@id"
     - name: basename
-      type: string?
+      type: string
```
mr-c: Why required? Doesn't match documentation below

Issue #183, opened 2016-01-28
[docs] tool description discovery


mr-c:
To be added to the documentation:

Programs that want to discover the CWL description for a tool should look in the following locations:

`/usr/share/cwl/`
`/usr/local/share/cwl/`
`$XDG_DATA_HOME/cwl/` (usually `$HOME/.local/share`)

`$XDF_DATA_HOME` is from the [XDG Base Directory Specification](http://standards.freedesktop.org/basedir-spec/basedir-spec-0.6.html)

Use cases:
1. Discovery of all installed tool descriptions.
2. Discovery of the tool description for a specific binary

mr-c:
Fixed in #246

PR #246, opened 2016-07-06
add instructions for discovering CWL documents on the local FS

mr-c:


mr-c:
Fix for https://github.com/common-workflow-language/common-workflow-language/issues/183

PR #278, opened 2016-07-26
v1.1 cwl-runner interface

mr-c:
Feedback is appreciated.

Are we keeping `--conformance-test`? What does it mean?


StarvingMarvin:
Re: conformance-test, I don't think it should be a part of cwl-runner, because you can always test command line generation of runners by running CWL jobs. Looking from a perspective of other CWL related tooling, like editors in our case, it might make sense to create separate interface / set of conformance tests, that doesn't require actual running.

mr-c:
@StarvingMarvin I agree. Any thoughts on alternative conformance testing?

StarvingMarvin:
@mr-c Well, it is a bit tricky. Depending on what the tool does it might have a limited understanding of CWL. Linter might be able to comprehend the structure of a document, but not know how to execute any expressions or to form command line. An expression tester might do exactly opposite: execute expressions with different values but not know how to parse much else beside expressions. Any centralized testing would have to have a mechanism of testing things in small chunks, with meaningful command line interface that shouldn't be bothersome to implement for someone that might possibly design a library or a GUI app that wouldn't otherwise have a command line interface. On a more "positive" note, it's not like there is plenty of tooling available, so we could tackle this problem once it become more specific.

In a meanwhile, we might just separate tests that are now handled with `--conformance-test` to a separate test-runner so that anything that that comes along and knows how to generate command, but not how to execute it can be tested with that.

tetron:
Agree with @starvingmarvin, --conformance-test is obsolete, it is already the case that all the draft-2, draft-3 and v1.0 tests have been updated to not use --conformance-test any more.

tetron:
FYI the --conformance-test flag meant to do a dry run and write out the command that would be run for a single tool.  It is a holdover from the draft-1 days.

tetron:
I think --basedir can also be removed from the generic interface.  It was used to manipulate the path prefix of inputs so that the --conformance-test command line would be the same, independent of the file system location of the inputs.

v1.0/cwl-runner.cwl
```diff
@@ -0,0 +1,78 @@
...
+ tmpdir-prefix:
+   type: string?
+   doc: |
+    Path prefix for temporary directories
+   inputBinding:
+    prefix: "--tmpdir-prefix"
+
+baseCommand: cwl-runner
+
+stdout: output-object.json
```
tetron: This should be the standard output object name, eg "cwl.output.json"

v1.0/cwl-runner.cwl
```diff
@@ -0,0 +1,78 @@
...
+
+outputs:
+  output-object:
+    type: stdout
```
tetron: This won't really do what we want, which is to return a structured object, but with "stdout: cwl.output.json" it will pick that up as the output object, which is what we want.

tetron: Err which means the outputs section can be empty, I think.

v1.0/cwl-runner.cwl
```diff
@@ -0,0 +1,62 @@
...
+    position: 2
+
+ no-container:
+   type: boolean?
+   doc: |
+    Do not execute jobs in a Docker container, even when listed as a Requirement
+   inputBinding:
+    prefix: "--no-container"
+
+ tmp-outdir-prefix:
```
brainstorm: At first glance it's confusing what's the actual difference between `tmp-outdir-prefix` and `tmp-prefix`... is one for input and the other for output? Just a docker/boot2docker hack? Why is not called i.e tmp-docker-prefix, etc... just naming concerns.

tetron:
If this is a minimal spec for cwl-runner, it should drop --no-container, --tmpdir-prefix and --tmp-output-prefix since they are implementation details that don't make sense for all systems (eg Arvados).

tetron:
fyi the difference between --tmpdir-prefix and --tmp-output-prefix is that the first is where the designated temporary directory should appear, the second is where intermediate workflow outputs are placed.  The temp directory is emphemeral and should go on fast local storage, outputs are persistent and need to go to durable stoage.

brainstorm:
Gotcha. Something like `tmpdir-ephemeral-prefix` (for lack of a better adjective) could help newcomers get the semantic/meaning of the flags quickly.

mr-c:
I've dropped the implementation specific options as suggested by @tetron , documented the exit code for unimplemented features, and clarified the outputs.

@tetron @jmchilton @StarvingMarvin Can I get a review from each of you (or your deputy) of this?

StarvingMarvin:
+1

v1.1.0-dev1/cwl-runner.cwl
```diff
@@ -0,0 +1,49 @@
...
+   type: File
+   doc: The input job document.
+   inputBinding:
+    position: 2
+
+baseCommand: cwl-runner
+
+outputs:
+  log: stderr
+  output_object_document: stdout  # in the CWL Output Object Document format
```
mr-c: `format: application/json`

LourensVeen:
Current cwltool can give an exit code 1 on error, besides the 0 and 33. Should that also be in there? And are there any exit codes for TemporaryFailure?

mr-c:
@LourensVeen Thank you for pointing that out.

Reviewing http://www.commonwl.org/v1.0/CommandLineTool.html#Execution we see that non-zero is a failure, by default. We can remind implementers of this with a comment, but I would prefer to not lock them into only returning `1` in the case of a permanent failure.

If the implementers would like an exit code for `cwl-runner` in the instance of a temporary failure I would be open to adding one.

tetron:
@LourensVeen this is a good point.  I think can think of the following states that would be useful to communicate with exit codes:

* loading / validation failure
* loaded but can't run due to use of unsupported feature (currently exit code 33)
* workflow temporary failure
* workflow permanent failure

LourensVeen:
@tetron: Interesting that you mention loading/validation errors separately, I hadn't considered that. I'd interpreted the spec as saying that such errors would yield a permanent failure. I'm not sure that it is useful to have a machine-readable distinction between the two; in both cases the user will probably have to do something about it.

@mr-c In general, error reporting is just as much part of your API as the command line arguments are. If you want people to be able to interface with cwltool from another program (or heck, write bash scripts using it), then it needs to be documented. And if you want people to be able to swap out different cwl-runners while doing so, then the exit codes need to be standardised, just like everything else.

I understand wanting to make life easy for implementers of cwl-runners, because you want CWL to be as widely supported as possible. But you're always going to have many more users than implementers. Making the users' lives harder in order to facilitate the implementers is on the whole not a good trade-off. And having to implement a particular set of exit codes seems a rather minor restriction compared to having to include a complete JavaScript interpreter.

Having said that, it turned out that for the REST-based CWL running service I'm building, the error status is not always available. I could work around it, but found it easier to just parse the "Final process status is ..." line in the stderr output. Of course that is probably even more non-standard, but it works for now, and gives me the detail I need.

v1.1.0-dev1/cwl-runner.cwl
```diff
@@ -0,0 +1,49 @@
+#!/usr/bin/env cwl-runner
+class: CommandLineTool
+cwlVersion: v1.1.0-dev1
+
+doc: |
+ Generic interface to run a Common Workflow Language tool or workflow from the
+ command line. To be implemented by each CWL compliant execution platform for
+ testing conformance to the standard and optionally for use by users.
+
+inputs:
```
mr-c: Also,
```
  version:
    type: boolean?
    doc: report the name & version, quit without further processing
    inputBinding: { prefix: "--version" }
```

PR #281, opened 2016-08-02
Move secondaryFiles, format and streamable to SchemaBase

pvanheus:
As discussed on Gitter with @tetron, secondaryFiles, format and streamable should belong to SchemaBase so that they can be used for compound types. For example:

```
  bam:
    type:
      - 'null'
      - type: array
        items: File
        inputBinding:
          prefix: "--bam"
          position: 6
        secondaryFiles:
          - ".bai"
    doc: |
      A bam file of unknown type; Pilon will scan it and attempt to classify it as one
      of the above bam types.
```

### DIFF
--- a/v1.0/Process.yml
+++ b/v1.0/Process.yml
@@ -314,15 +314,6 @@ $graph:
       jsonldPredicate: "rdfs:label"
       doc: "A short, human-readable label of this object."
 
-
-- name: Parameter
-  type: record
-  extends: SchemaBase
-  abstract: true
-  doc: |
-    Define an input or output parameter to a process.
-
-  fields:
     - name: secondaryFiles
       type:
         - "null"
@@ -380,6 +371,15 @@ $graph:
         indicate whether it is valid to stream file contents using a named
         pipe.  Default: `false`.
 
+
+- name: Parameter
+  type: record
+  extends: SchemaBase
+  abstract: true
+  doc: |
+    Define an input or output parameter to a process.
+
+  fields:
     - name: doc
       type:
         - string?


mr-c:
Unfortunately this means that types unrelated to `File` get `secondaryFiles`, `format`, and `streamable` as properties: `CommandInputRecordSchema`, `CommandInputEnumSchema`, `CommandOutputRecordSchema`, `CommandOutputEnumSchema`, `InputRecordSchema`, and `InputEnumSchema`,

The documentation for all three of those properties starts with "Only valid when type: File or is an array of items: File."

Note: `CommandOutputArraySchema` does not get the desired properties in this PR.

I agree that it would be good to add these properties to `InputArraySchema`, and `OutputArraySchema` which will flow to `CommandInputArraySchema`, `CommandOutputArraySchema` automatically.

@pvanheus I've opened a PR that does just that: https://github.com/pvanheus/common-workflow-language/pull/1

mr-c:
By the way @pvanheus , very impressive first take on the schema salad!

tetron:
@pvanheus @mr-c I think we might want to do this a little bit differently, I agree SchemaBase is the wrong place for these fields, but they actually need to be in each of `Parameter`, `InputArraySchema`, `InputRecordField`, `OutputArraySchema` and `OutputRecordField`.

Instead of copying the fields, there should be an abstract base that these can all inherit from.

mr-c:
@tetron I'm okay with a new base type, but `Parameter` is also the wrong place as `SchemaBase` extends `Parameter` and the same problem would be created.

mr-c:
Strike that, I had the relationship between Parameter & SchemaBase backwards. I'll go create the new base type in my PR to this PR.

tetron:
Right, `Parameter` inherits from `SchemaBase` but `SchemaBase` is also a base for other types for which the file-related fields are not appropriate (record and enum types specifically) which is why the correct solution is to introduce a new base type and add that to the `extends` declaration of the types I listed above.  (Fortunately, schema salad supports multiple inheritance)

mr-c:
These changes have been moved to #305, discussion can continue there

PR #282, opened 2016-08-02
Add test for command line inputBinding of Directory input parameter.

tetron:


tetron:
See https://github.com/common-workflow-language/cwltool/pull/144 for related cwltool bugfix

PR #288, opened 2016-08-15
Clarify logo license

mr-c:
to match https://github.com/common-workflow-language/logo/blob/master/LICENSE.md

### DIFF
--- a/LICENSE.txt
+++ b/LICENSE.txt
@@ -1,3 +1,7 @@
+All files are released under the Apache License, version 2.0, except
+"master/site/CWL-Logo-Header.png" which is released under the GNU Lesser
+General Public License, version 3 or any later version, or, at your
+option, of the Creative Commons Attribution-ShareAlike 3.0 Unported License.
 
                                  Apache License
                            Version 2.0, January 2004


stain:
I don't think you should add that to the top of `license.txt` as it then looks like a modified Apache license. Perhaps rather add a `--------------------------------` line in the end and say the licenses of the logo separately.

From https://www.gnu.org/licenses/lgpl-3.0#section4 you would also need to:

> b) Accompany the Combined Work with a copy of the GNU GPL and this license document.

E.g. add both the GPL and LGPL license text.  (This should be true also for the /logo repository). You could add them as files next to the logo, e.g. `site/CWL-Logo-Header-LICENSE-LGPL.txt` etc.

> 1) Use a suitable shared library mechanism for linking with the Library. A suitable mechanism is one that (a) uses at run time a copy of the Library already present on the user's computer system, and (b) will operate properly with a modified version of the Library that is interface-compatible with the Linked Version.

Being in a separate file is pretty interface-compatible :)

And as the source code (*.ai) is not included, this is the _object form_, so [section 6 of GPL 3](https://www.gnu.org/licenses/gpl-3.0#section6) applies:

> d) Convey the object code by offering access from a designated place (gratis or for a charge), and offer equivalent access to the Corresponding Source in the same way through the same place at no further charge. You need not require recipients to copy the Corresponding Source along with the object code. If the place to copy the object code is a network server, the Corresponding Source may be on a different server (operated by you or a third party) that supports equivalent copying facilities, provided you maintain clear directions next to the object code saying where to find the Corresponding Source. Regardless of what server hosts the Corresponding Source, you remain obligated to ensure that it is available for as long as needed to satisfy these requirements.

So the URL back to the source code in https://github.com/common-workflow-language/logo/ must be provided.

PR #291, opened 2016-08-19
Adding docker hint to conformance test workflow

kellrott:
Python wasn't install on the system the conformance test was being run on, so the hint for a python docker container was required to get the job to run.


mr-c:
Thanks!

PR #292, opened 2016-08-22
add test for accessing array of secondaryFiles in docker

jeremiahsavage:
Add a test for accessing an array of secondaryFiles within docker. This test fails on Master. I am creating a PR to allow it to pass.


jeremiahsavage:
This PR will allow the test to pass https://github.com/common-workflow-language/cwltool/pull/170

tetron:
Thanks for posting this.  I've updated the test in #324

mr-c:
Fixed in #324

PR #304, opened 2016-09-07
document and test requirements in input job objects

mr-c:


mr-c:
@tetron @jmchilton @StarvingMarvin Can I get a review from each of you (or your deputy) of this?

mr-c:
Relevant content is at https://github.com/common-workflow-language/common-workflow-language/pull/304/files#diff-0b64c4b6bf9a00f68298cf91f64bf23aR212

Note the tests and how conflicts are resolved -- the `requirement` in the input job object will override a requirement of the same type in the Workflow/CommandLineTool -- even if that requirement allows for subparts.

jmchilton:
Language change seems fine for an implementation like cwltool - insecure* and file-centric - but I'm in principle a -0.5 on this and any other comments like this related to the input document until inputs are spun out into a separate complimentary document. I think the core of the specification should be more secure and should not allow this.

My concerns regarding coercing the format of inputs would not be helped by the following change, but at least it would address the security concerns I have I think:

Instead of:

```
Requirements can be specified in the input object document as an array of
requirements under the field name `cwl:requirements`. They are combined with
any requirements present in the corresponding Process as if they were specified
there.
```

Say:

```
Optionally, implementations may allow requirements to be specified in the input object document
as an array of requirements under the field name `cwl:requirements`. If implementations allow this
such requirements should be combined with any requirements present in the corresponding Process
as if they were specified there.
```

* Security being defined unfairly and narrowly as an implementation should be able to define a set of CWL operations (tools or tools & workflows) it allows and then be confident that end-users can supply inputs that don't break desired command-line parsing, don't specify external docker containers, etc...

mr-c:
Moved to https://github.com/common-workflow-language/cwl-v1.1/pull/19

PR #309, opened 2016-09-08
Test for using $mixin 

tetron:
In this example, to simplify redundant "inputs:" section in subworkflow.


tetron:
Note, requires a schema salad fix.

karl616:
I set up a pipeliine today that wished that this problem was solved.
Is there another solution to have something like a global settings file?

tom-tan:
Can we close this issue?
`$mixin` was already removed in [Schema Salad v1.2](https://www.commonwl.org/v1.2/SchemaSalad.html#Introduction_to_v1.1).

mr-c:
Thank you @tom-tan  for catching this!

PR #314, opened 2016-09-15
Add test for requirements/hints on workflow steps.

tetron:


mr-c:
Can these be added to v1.1 as well?

Pe 15 sept. 2016 8:04 p.m., "Peter Amstutz" <notifications@github.com> a
scris:

PR #319, opened 2016-10-07
Add additional test for cwl.output.json behavior.

tetron:


tetron:
Test for https://github.com/common-workflow-language/cwltool/pull/206

Issue #326, opened 2016-10-22
wrong version links in documentation


sb10:
http://www.commonwl.org/v1.0/Workflow.html section 2.2 has a link to https://github.com/common-workflow-language/common-workflow-language/tree/master/draft-4, which I presume should be https://github.com/common-workflow-language/common-workflow-language/tree/master/v1.0/salad ?

The README.rst file in there has a Documentation section that links to http://www.commonwl.org/draft-3/SchemaSalad.html instead of, I presume, http://www.commonwl.org/v1.0/SchemaSalad.html.

I haven't looked beyond these 2 examples for similar errors, but perhaps a quick grep for "draft" in all v1.0 files will identify everything that needs to be fixed.

PR #328, opened 2016-10-25
Fixing mentions to draft3/4

StarvingMarvin:
Related to #326.

I haven't gone through v1.1-dev because hopefully, we'll change those links again in a week or so :)


mr-c:
Thanks!

PR #339, opened 2016-11-09
Add test for dockerOutputDirectory option of DockerRequirement.

tetron:
Link to the pertinent part of the specification: http://www.commonwl.org/v1.0/CommandLineTool.html#DockerRequirement


mr-c:
Thanks

PR #343, opened 2016-11-14
Clarify SoftwareRequirement spec fields with better examples

mr-c:
As it is just a clarification I can merge this with 1.0 or leave it for 1.1.0; feedback is very welcome.

In summary: no specification of what software corresponds to a particular CWL description is required, if one does specify then only a name is required. A portable identifier is recommended, especially if the name conflicts with other packages or is inconsistently specified. That identifier can be mapped to specific mechanisms by the CWL implementation. Additionally allowed are identifiers pointing at specific mechanism from the generic (bioconda, Debian) to the site specific (Environment Modules).

Remember that today, in many CWL implementations including `cwltool`, users can specify Requirements (such as a site specific SoftwareRequirement) in the user input job object using `cwl:requirements`. Starting with v1.1.0  all CWL implementations that accept user created input job objects will be required to support the `cwl:requirements` field.

### DIFF
--- a/site/cwlsite.cwl
+++ b/site/cwlsite.cwl
@@ -26,7 +26,7 @@ inputs:
           rdfs_target: string
   brandimg: File
   empty:
-    type: string
+    type: string?
     default: ""
 
 outputs:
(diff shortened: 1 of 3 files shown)


jmchilton:
Thanks - this looks really good to me.

tetron:
Merged into #410

PR #376, opened 2017-01-26
long overdue linking to open-stand.org

mr-c:


mr-c:
Jenkins, test this please

mr-c:
Thank you @stain 

We recently agreed to slow down spec changes and use patch versioning for every change, no matter how small.

PR #382, opened 2017-02-08
Test if a writable input directory is recursivly copied and writable

gijzelaerr:
This is to test the upcoming patch release of the v1.0 series to clarify https://github.com/common-workflow-language/cwltool/issues/282

do not merge yet

### DIFF
--- a/v1.0/conformance_test_v1.0.yaml
+++ b/v1.0/conformance_test_v1.0.yaml
@@ -983,7 +983,7 @@
   doc: Test hints with $import
 
 - job: v1.0/recursive-input-directory.json
-  output:
+  output: {
    output_dir: {
         "basename": "work_dir",
         "class": "Directory",
@@ -1023,17 +1023,17 @@
                 "location": "work_dir/e",
                 "size": 0
             },
-            {
-                "checksum": "sha1$da39a3ee5e6b4b0d3255bfef95601890afd80709",
-                "basename": "testresult",
-                "location": "work_dir/testresult",
-                "class": "File",
-                "size": 0
-            }
         ],
         "location": "work_dir",
+    },
+    "test_result": {
+        "basename": "output.txt",
+        "checksum": "sha1$da39a3ee5e6b4b0d3255bfef95601890afd80709",
+        "class": "File",
+        "location": "output.txt",
+        "size": 0
     }
-
+}
 
 
 
--- a/v1.0/conformance_test_v1.0.yaml
+++ b/v1.0/conformance_test_v1.0.yaml
@@ -981,3 +981,25 @@
       size: 15
   tool: v1.0/imported-hint.cwl
   doc: Test hints with $import
+
+- job: v1.0/recursive-input-directory.json
+  output:
+    "outdir": {
+        "class": "Directory",
+        "listing": [
+            {
+              "class": "File",
+              "location": "hello.txt",
+                "checksum": "sha1$47a013e660d408619d894b20806b1d5086aab03b",
+                "size": 13
+            },
+            {
+              "class": "File",
+              "location": "goodbye.txt",
+                "checksum": "sha1$dd0a4c4c49ba43004d6611771972b6cf969c1c01",
+                "size": 24
+            }
+        ],
+    }
+  tool: v1.0/recursive-input-directory.yml
+  doc: Test if a writable input directory is recursivly copied and writable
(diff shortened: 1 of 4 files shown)


v1.0/v1.0/recursive-input-directory.yml
```diff
@@ -0,0 +1,18 @@
...
+        writable: true
+baseCommand: "touch"
+arguments: [$(inputs.input_dir.path + "/e")]
+inputs:
+  input_dir: Directory
+outputs:
+  output_dir:
+    type: Directory
+    outputBinding:
+      glob: $(inputs.input_dir)
```
mr-c: `glob: $(inputs.input_dir.path)`
which won't work because `glob` can't start with a `/` -- use `glob: word_dir` instead

v1.0/v1.0/recursive-input-directory.yml
```diff
@@ -0,0 +1,18 @@
...
+class: CommandLineTool
+requirements:
+  - class: InlineJavascriptRequirement
+  - class: InitialWorkDirRequirement
+    listing:
+      - entry: $(inputs.input_dir)
+        entryname: work_dir
+        writable: true
+baseCommand: "touch"
+arguments: [$(inputs.input_dir.path + "/e")]
```
mr-c: I would use `baseCommand: [ touch, work_dir/e ]` instead of these two lines

v1.0/v1.0/recursive-input-directory.yml
```diff
@@ -0,0 +1,18 @@
+cwlVersion: v1.0
```
mr-c: This file should have the `.cwl` extension

v1.0/conformance_test_v1.0.yaml
```diff
@@ -981,3 +981,25 @@
       size: 15
   tool: v1.0/imported-hint.cwl
   doc: Test hints with $import
+
+- job: v1.0/recursive-input-directory.json
+  output:
+    "outdir": {
```
mr-c: YAML please :-)

mr-c:
Jenkins, add to test list

v1.0/conformance_test_v1.0.yaml
gijzelaerr: sorry, what do you mean?

gijzelaerr: you mean the `"outdir": {` is not yaml syntax? The same syntax is used in the same file?

gijzelaerr:
ok now I have a passing test, but that is not what I want. I want to make sure the files in work_dir are a recursvie copy from the input, and not a symlink.... right?

mr-c:
@gijzelaerr Do you want to test writability, or if the directory was symlinked in or not?

If the later, try this:

```
cwlVersion: v1.0
class: CommandLineTool
requirements:
  InitialWorkDirRequirement:
    listing:
      - entry: $(inputs.input_dir)
        entryname: work_dir
        writable: true
baseCommand:
 - bash
 - '-c'
 - 'if [ -L work_dir ]; then echo is symlink; else echo not symlink; fi'
inputs:
  input_dir: Directory
outputs:
  result: stdout
```

gijzelaerr:
ok done. now I probably should check for the output in `conformance_test_v1.0.yaml` right to make sure the CWL script didn't echo something... but it is not obvious to me how? It looks like I can only check for expected output files and not expected output on the console?

mr-c:
@gijzelaerr In the `conformance_test_v1.0.yaml` we can specify the correct outputs, in this instance it is a  File with a particular sha1sum and size. So make the correct file and update the sha1sum and size to match.

gijzelaerr:
Yes, but i'm outputting to the console here, not to a file. It doesn't look like the conformance test supports checking console output? If so then probably I should write the failing tests to a file and update the expected output sha1 conformance test with the sha1 of an empty file.

mr-c:
You can capture the stdout output into a file: http://www.commonwl.org/v1.0/UserGuide.html#Capturing_a_tool's_standard_output_stream

gijzelaerr:
ok I think I do it correctly now. So the next step would get 2 reference implementations working?

mr-c:
@gijzelaerr just a PR to fix `cwltool` is good enough, thanks!

mr-c:
Now that the change is implemented in `cwltool` I'll take this branch over to add the clarification to the spec

tetron:
Merged into #410

PR #383, opened 2017-02-09
Use v1.0 style id's in inputs/outputs example

andrewjesaitis:
Super minor issue, but just saw that the old `id` syntax was still in the v1.0 user guide. Also fixed it in the 1.1-dev1 folder, but I'm not sure if this will be valid 1.1 syntax. Thanks!


mr-c:
Jenkins, add to test list

PR #395, opened 2017-02-28
Make the "dir5.cwl" pass if the implementation uses symlinks.

tetron:


tetron:
Jenkins, test this please

PR #400, opened 2017-03-02
Sync v1.1 conformance test suite to changes in v1.0 suite.

tetron:


v1.1.0-dev1/conformance_test_v1.1.0-dev1.yaml
```diff
@@ -87,7 +87,7 @@
       size: 84
       location: error.txt
 
-- job:
+- job: v1.1.0-dev1/empty.json
   tool: v1.1.0-dev1/egrep-stderr-shortcut.cwl
   doc: Test command line with stderr redirection, brief syntax
```
mr-c: This test is failing: 
```
{
    "output_file": {
        "checksum": "sha1$6039b7cf804ccce91d7ca4fabca0a5ee9cb583ff", 
        "basename": "c8c06346-da20-4a05-8fcc-e2426a42b007", 
        "location": "file:///tmp/tmpD4Cit2/c8c06346-da20-4a05-8fcc-e2426a42b007", 
        "path": "/tmp/tmpD4Cit2/c8c06346-da20-4a05-8fcc-e2426a42b007", 
        "class": "File", 
        "size": 338
    }
}
```

tetron: I don't know why this test is failing.  However, these tests are a bit unstable in that they rely on egrep printing out a specific usage message on stderr.  I am updating these tests to use "echo 2>&1" to print a message to stderr.

mr-c:
Hrmm, still failing on bunny. I've asked for their help: https://github.com/rabix/bunny/issues/183

tetron:
@mr-c you can call /bin/sh -c directly (the way it was originally written) or use ShellCommandRequirement but using both doesn't do what you want.

If bunny fails the tests using /bin/sh -c but passes with ShellCommandRequirement, they might have a quoting problem.  So we might want to add an additional test specifically check that shell metacharacters don't break regular arguments.

mr-c:
@tetron Right, sorry. Now down to two Bunny failures

mr-c:
Ah, but they are different than the other failures -- so we are probably ok!

PR #401, opened 2017-03-03
Add test of arguments containing shell directives 

tetron:
Add test to ensure that arguments containing shell directives are not interpreted and that `shellQuote: false` has no effect when ShellCommandRequirement is not in effect.


tetron:
Merged into #410

PR #403, opened 2017-03-07
Remove superfluous resource requirement

bosonogi:
"General test of command line generation" should not require a minimum of four cores.

Briefly discussed with @StarvingMarvin

### DIFF
--- a/v1.0/v1.0/bwa-mem-tool.cwl
+++ b/v1.0/v1.0/bwa-mem-tool.cwl
@@ -5,8 +5,6 @@ cwlVersion: v1.0
 class: CommandLineTool
 
 hints:
-  - class: ResourceRequirement
-    coresMin: 4
   - class: DockerRequirement
     dockerPull: python
 
--- a/v1.1.0-dev1/v1.1.0-dev1/bwa-mem-tool.cwl
+++ b/v1.1.0-dev1/v1.1.0-dev1/bwa-mem-tool.cwl
@@ -5,8 +5,6 @@ cwlVersion: v1.1.0-dev1
 class: CommandLineTool
 
 hints:
-  - class: ResourceRequirement
-    coresMin: 4
   - class: DockerRequirement
     dockerPull: python
 


v1.0/v1.0/bwa-mem-tool.cwl
```diff
@@ -5,8 +5,6 @@ cwlVersion: v1.0
 class: CommandLineTool
 
 hints:
-  - class: ResourceRequirement
-    coresMin: 4
```
mr-c: Can we replace this with other tests that provide minimal coverage of `ResourceRequirement`?

StarvingMarvin: @mr-c I agree, using data from resource requirement is part of the test, but it should be lower, because it prevents running conformance tests on 2-core amazon instances, which is silly.

mr-c: I am totally happy for the specified value to be super minimal :-)

bosonogi: @StarvingMarvin What guaranties would we have that the data from resource requirements is actually honored?

Should we make a new category of tests, of jobs that should fail? Then we could make some ridiculous requirement for a simple tool, like `coresMin: 9000`, and be more certain that a cwl runner is taking the requirement into account.

mr-c: For the bare minimum of test coverage of the spec it would be good to have each part of ResourceRequirement used in a conformance test, even if we don't verify it.

:+1: for expected failing tests (and a third category of optional tests (for optional features) can be added at the same)

StarvingMarvin: @bosonogi Question about honoring requirements is a different test. This one tests if the value will be propagated to the command line.

bosonogi: @StarvingMarvin Oh, you mean the `$(runtime.cores)` bit? Sorry I missed that before.

And yes, I was aware that what I was suggesting would be a different test.

tetron:
Addressed in #410 by adjusting resource request from 4 cores to 2 cores (as discussed on gitter).

PR #407, opened 2017-03-08
fix support link

mr-c:


tetron:
Not so simple.  The problem is that github generates different default anchor links from the schema salad makedoc, so either you fix commonwl.org and break the github README or vice versa.

Maybe you can embed an explicit anchor that has a stable id?

mr-c:
@tetron Good catch, I fixed that; thanks!

PR #408, opened 2017-03-08
changes to enable nolinkcheck in case of default field, related to cwltool issue#2

kapilkd13:
This change make default field to show warning instead of throwing error, by enabling nolinkcheck flag
Solves https://github.com/common-workflow-language/cwltool/issues/2


mr-c:
Jenkins, okay to test

kapilkd13:
Hi @mr-c Can you guide me on why the conformance tests are failing on this PR. I am little surprised that tests are failing for draft 2 and 3, as I made changes on v1 and v1-dev.

mr-c:
For this PR, all the conformance test with the reference implementation are passing. Arvados/arvbox has been having troubles for a long time, so that isn't a particular concern.

Looks like the development release of Rabix has had an uneven history with the conformance tests lately.

Lets test again to see what is the current status.

Jenkins, test this please.

mr-c:
It would be great if this change was accompanied by a new test case that only passes when this fix is present.

kapilkd13:
I will add a test case for this change. Thanks

kapilkd13:
Hi @mr-c I made a PR for this change, currently it should give error. After this merge it should work. Would you suggest any change in this unit test.

mr-c:
@kapilkd13 Looks good, the tests fail in the expected manner.

Once you add those commits to this PR I'll merge this. Thanks!

kapilkd13:
@mr-c I have added conformance test regarding this PR.

mr-c:
Jenkins, test this please

v1.0/v1.0/default_path_job.yml
```diff
@@ -0,0 +1,3 @@
+file1:
+  class: File
+  path: hello.txt
```
jmchilton: A lot of these paths seemed to be switched to locations in these test cases but in light of this test case I'm wondering if the "may" in the following sentence should be a "must"?

> As a special case, if the path field is provided but the location field is not, an implementation may assign the value of the path field to location, and remove the path field.

PR #417, opened 2017-03-27
Fix identifier conflicts (checked by cwltool for the first time) 

tetron:
Give ids to embedded tools and workflows.


mr-c:
@tetron If `id`s are required for sub-processes we need to be more explicit about that in the spec.

Ping @StarvingMarvin

StarvingMarvin:
Presence (or absence) of these ids won't affect bunny in any way, but it would be nice to see more detailed explanation of what should it mean spec-wise.

PR #410, opened 2017-03-12
v1.0.1 (errata) 

tetron:
Criteria to be considered for v1.0.1: must not affect syntax of v1.0 documents.  Existing v1.0 documents must remain valid.  May affect behavior for documents that previously relied on undefined or underspecified behavior.

Needs tests:

- [x] Test for location of files in InitialWorkDir
- [x] Test for recursively writable: https://github.com/common-workflow-language/common-workflow-language/pull/382
- [x] Test of workflow in compound document
- [x] Test that `default` applies when source is `null`
- [x] Test that missing parameters are assigned `null` by default
- [x] Conformance test for nameroot and nameext #454

Documentation todos:

- [x] Merge discussion from #348 
- [x] Merge updated schema-salad documentation to include documentation of mapping transforms and type DSL
- [x] Scatter over empty list is empty
- [x] Clarify interpretation of secondaryFiles on inputs (all files are expected to be present, no optionals)
- [x] What to do if "format" field on output is an array: #482 
- [ ] clarify the subject of every imperative sentences: for each sentence, is this the user's job or the implementation's job? @mr-c

Additional release checklist:

- [ ] Get sign off from SBG, other stakeholders
- [ ] Update docs
  - [x] Add a detailed "What's changed" section for `v1.0.1`

Post-merge:

- [ ] Add github release
- [ ] Upload to figshare, get updated DOI

PR #435, opened 2017-05-05
Empty scatter

StarvingMarvin:
Specification clarification regarding scattering on empty lists, and accompanying conformance tests.


tetron:
Merged into #410

PR #451, opened 2017-06-06
Add test for MultipleInputFeatureRequirement

michael-kotliar:
If the source has a type set as a list of two or more items, it should be added to the result array exactly once and not as `cwltool` mistakenly does: the number of times as the length of its type list


mr-c:
Jenkins, test this please

mr-c:
@tetron @StarvingMarvin I am going to merge this new conformance test in 24 hours if there is no objection

tetron:
Merged into #410

PR #454, opened 2017-06-09
Add conformance test for nameroot and nameext

undu:
Right now cwltool doesn't generate nor nameroot nor nameext at execution time from basename of an input File.

In the specification it states that they should be, now we have a test for it!


mr-c:
Jenkins, add to testlist

mr-c:
Jenkins, test this please

tetron:
Merged into #410

PR #460, opened 2017-06-21
Add tests for $(null) and .length on arrays.

tetron:


tetron:
Thanks.  Will hold off merge after publishing an updated cwltool to avoid breaking builds.

PR #462, opened 2017-06-22
Add conformance test for static_checker

michael-kotliar:
Added `count-lines11-wf.cwl`

## Expected Behavior 
Workflow input
```json
file1:
  type:
    - type: array
      item: File
```
should be exactly the same as
```json
file1:
  type: File[]
```
## Actual Behavior
`static_checker` mistakenly fails with source -> sink mismatch
```
Tool definition failed validation:
count-lines11-wf.cwl:9:5: Source 'file1' of type [{"type": "array", "items": "File"}] is incompatible
count-lines11-wf.cwl:25:7:   with sink 'file1' of type {"type": "array", "items": "File"}
count-lines11-wf.cwl:25:7:   sink has linkMerge method merge_flattened
```
## Your Environment
* cwltool version: 1.0.20170622090721

### DIFF
--- a/v1.0/conformance_test_v1.0.yaml
+++ b/v1.0/conformance_test_v1.0.yaml
@@ -981,3 +981,11 @@
       size: 15
   tool: v1.0/imported-hint.cwl
   doc: Test hints with $import
+
+- job: v1.0/count-lines6-job.json
+  output:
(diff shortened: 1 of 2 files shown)


tetron:
Merged into #410

PR #469, opened 2017-07-04
Add support for cwltest -j to run_test.sh

tetron:


tetron:
Propagate `run_test.sh -j4` to `cwltest -j4` to run 4 test jobs in parallel.

PR #475, opened 2017-07-09
Container correction: Adding missing Docker container in conformance tests

kapilkd13:
Some of the workflows/tools in conformance tests use utilities like python, bash. Currently assumes that they would be available by default on running machine. But as we are adding support for windows in which we would run these workflows in a docker container, we need to provide these containers.


mr-c:
@kapilkd13 Thanks!

PR #481, opened 2017-07-19
Container correction: adding docker containers as hints

kapilkd13:
These conformance tests uses `rev` to reverse, this command is not posix compliant. Therefore on default alpine container on windows it behaves differently then on ubuntu/debian image. Therefore adding `debian :wheezy` docker container as hints


kapilkd13:
Hi @mr-c Since we have changed default docker Container for windows. I think we should merge this.

mr-c:
@kapilkd13 Thanks!

PR #502, opened 2017-08-15
Add test that names are preserved with reverse mapping of output.  Al…

tetron:
…so test

that objects with 'id' field are legal in the 'out' of a workflow step.


mr-c:
Rabix Bunny currently fails "Test support for reading cwl.output.json in a workflow.", sending it to them for review

PR #503, opened 2017-08-15
missing back-tick causes misformatting

mccalluc:


mr-c:
Thank you @mccalluc , FYI, this user guide is for an outdated version of CWL -- in fact we are about to launch http://www.commonwl.org/user_guide/ as the replacement

PR #506, opened 2017-08-16
Broken value from

tetron:


mr-c:
@tetron Am I missing something besides the renames and moving with the list of tests?

tetron:
Nope, it's just the rename, I realized after merging the initial test that we already had a set of related tests named scatter-valuefrom and I wanted it to make it part of that group.

PR #507, opened 2017-08-17
Test file literal as input without Docker

mr-c:
Exact same test as 81, but without the Docker hint.

Fails with latest cwltool


v1.0/v1.0/cat3-nodocker.cwl
```diff
@@ -0,0 +1,16 @@
+#!/usr/bin/env cwl-runner
+class: CommandLineTool
+cwlVersion: v1.0
+doc: "Print the contents of a file to stdout using 'cat' running in a docker container."
```
tetron: docstring is wrong

PR #511, opened 2017-09-08
[MRG] remove old user guide, link to the new one

mr-c:


mr-c:
Many, many thanks to @tetron for writing the original user guide on which the new one is heavily based!

PR #520, opened 2017-09-13
Update link for Airflow

michael-kotliar:
Update link to point to correct
cwl-airflow repository.
Update description


mr-c:
@michael-kotliar Cool! Guess I should update ci.commonwl.org as well?

PR #492, opened 2017-08-07
Oodles of new conformance tests

jmchilton:
There was a bunch of complexity related to simple input and output, default values, optional values, Any, union types, etc... to workflows and tools in Galaxy. This adds a bunch of conformance tests that isolates these basic input and output concerns.

There were also test cases for subworkflow scattering - this adds a couple of those.

PR #523, opened 2017-09-18
Tests demo-ing specification of secondaryFiles in subdirectories of the job input document.

jmchilton:
Other secondaryFile input object test specified files in the same directory as the job input document. To me it was not obvious if the locatin/path of the secondary inputs should have been relative to their parent input or relative to the job itself.

Unfortunately cwltool just picks up secondaryFiles as needed without them being specified, so I'm not absolutely certain the behavior is the intended behavior but it is my best guess.

### DIFF
--- a/v1.0/conformance_test_v1.0.yaml
+++ b/v1.0/conformance_test_v1.0.yaml
@@ -895,6 +895,30 @@
   tool: v1.0/dir4.cwl
   doc: Test directories in secondaryFiles
 
+- job: v1.0/dir4-subdir-1-job.yml
+  output: {
+    "outlist": {
+      "checksum": "sha1$9d9bc8f5252d39274b5dfbac64216c6e888f5dfc",
+      "size": 14,
+      "location": "output.txt",
+      "class": "File"
+    }
+  }
+  tool: v1.0/dir4.cwl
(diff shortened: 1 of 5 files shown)


tetron:
Inferred secondary files should be relative to the primary file path or location.

jmchilton:
@tetron So it is definitely a bug in cwltool that on this branch the job configuration:

```
inf:
  class: File
  location: subdirsecondaries/hello.py
  secondaryFiles:
    - class: File
      location: hello.tar
    - class: Directory
      basename: xtestdir
      location: testdir
```

finds the testdir in ``v1.0`` instead of ``v1.0/subdirectories``?

tetron:
> Update: Or is that only for "inferred" secondary files instead of explicit, specified secondary files?

Yes.  Explicit files are resolved relative to the document base, inferred files are resolved relative to the primary.

jmchilton:
Okay these tests should confirm the first half of that then and this is cwltool's behavior. I still have issues with inferred secondary files - but those are for a different PR I suppose.

jmchilton:
Merged as part of https://github.com/common-workflow-language/common-workflow-language/pull/492/files I believe.

PR #528, opened 2017-10-01
Update Debian Docker image used in the v1.0 tests

cure:
Run all v1.0 tests that use a Debian Docker image with the debian:stretch-slim image instead of debian:8 or debian:wheezy.

The debian:stretch-slim is smaller than either of the others.


mr-c:
Thank you @cure !

PR #558, opened 2017-11-29
Move files under v1.0/ for conformance test

manabuishii:
examples/arguments-job.yml
examples/linkfile.cwl
are used for conformance test

But they are not exists under v1.0/ directory
It should be better all files exists under v1.0
 for conformance test


mr-c:
Thanks!

PR #577, opened 2017-12-20
test for POSIX glob behavior

mr-c:
Tests https://github.com/common-workflow-language/cwltool/pull/587


manabuishii:
I think not only test with 0 byte files but also at least one file has more than 0 to confirm body of file is successfully available.

mr-c:
@manabuishii I think we test for file availability in other conformance tests

manabuishii:
OK I understand

mr-c:
Jenkins, test this please

PR #585, opened 2018-01-08
Adding conformance test for workflow default override

kellrott:
Making sure workflow default overrides the tool default. 
Fixes issue https://github.com/common-workflow-language/cwltool/issues/592
cwltool issue fixed in https://github.com/common-workflow-language/cwltool/pull/595


kellrott:
The automated testing errors seem to be showing up on master as well, so this PR probably didn't create them.

PR #599, opened 2018-01-18
Using requirements instead of hints

manabuishii:
In terms of `doc` comment in conformance_test_v1.0.yaml
These 3 files using for Docker.
These 3 tests are needed to use `requirements`.

But `cat3-tool.cwl` is used as
I don't know how to effect this.
I think following test using another file.

```
- job: v1.0/file-literal.yml
  output:
    output_file:
      class: File
      checksum: sha1$d0e04ff6c413c7d57f9a0ca0a33cd3ab52e2dd9c
      location: output.txt
      size: 18
  tool: v1.0/cat3-tool.cwl
  doc: Test file literal as input
```

### DIFF
--- a/v1.0/conformance_test_v1.0.yaml
+++ b/v1.0/conformance_test_v1.0.yaml
@@ -72,11 +72,11 @@
       checksum: sha1$47a013e660d408619d894b20806b1d5086aab03b
       location: output.txt
       size: 13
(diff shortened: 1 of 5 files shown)


manabuishii:
And this PR force to use Docker at conformance test .

PR #610, opened 2018-01-27
Test for empty not-null boolean inputBinding

### LINKED ISSUE #604: explicit conformance test for boolean with empty (but not null) inputbinding

anton-khodak:
closes #604 
cwltool currently fails this test: appends "True" to a command-line instead of ignoring such an input


mr-c:
Thanks! We might backport this test to v1.0 as well. @tetron, yes?

mr-c:
Upon reflection this test should be added to `v1.0` as well

PR #613, opened 2018-01-27
Replace `wc` with `cat` to improve test's portability

### LINKED ISSUE #79: wc test failing on different platforms

anton-khodak:
Closes #79

### DIFF
--- a/v1.0/conformance_test_v1.0.yaml
+++ b/v1.0/conformance_test_v1.0.yaml
@@ -152,14 +152,14 @@
     output: 2
   doc: Testing the string 'null' does not trip up an Any without a default value.
 
-- job: v1.0/wc-job.json
+- job: v1.0/cat-job.json
   output:
     output:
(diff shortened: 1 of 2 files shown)


v1.0/conformance_test_v1.0.yaml
```diff
@@ -152,14 +152,14 @@
     output: 2
   doc: Testing the string 'null' does not trip up an Any without a default value.
 
-- job: v1.0/wc-job.json
+- job: v1.0/cat-job.json
```
mr-c: What would be the impact of removing the `wc-*` files?

anton-khodak: They are necessary for a bunch of tests, so I decided not to remove them. Those tests seem to be stable across platforms.

PR #615, opened 2018-01-28
modify template to avoid the "with" codepath, which violates strict mode

mr-c:


mr-c:
Thank you @tetron ! More context is at http://underscorejs.org/#template

PR #624, opened 2018-02-02
Add test for trailing whitespace in file entries of InitialWorkDir

anton-khodak:
Inspiration: https://github.com/common-workflow-language/cwltool/issues/247

### DIFF
--- a/v1.0/conformance_test_v1.0.yaml
+++ b/v1.0/conformance_test_v1.0.yaml
@@ -462,6 +462,17 @@
   doc: |
     Test InitialWorkDirRequirement with expression in filename.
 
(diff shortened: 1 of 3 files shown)


v1.0/conformance_test_v1.0.yaml
```diff
@@ -462,6 +462,12 @@
   doc: |
     Test InitialWorkDirRequirement with expression in filename.
 
+- job: v1.0/string-job.json
+  output:
+    out: "CONFIGVAR=hello\n"
```
mr-c: There's not `\n` in YAML :-) How about a regular `File` output and a checksum?

Issue #627, opened 2018-02-06
missing conformance tests for {success,temporaryFail,permanentFail}Codes


mr-c:
Needed for v1.0.x and v1.1

PR #628, opened 2018-02-06
test for alternative success exit codes

mr-c:
Initial fix for #627

### DIFF
--- a/v1.0/v1.0/exit-success.cwl
+++ b/v1.0/v1.0/exit-success.cwl
@@ -3,7 +3,7 @@ cwlVersion: v1.0
 class: CommandLineTool
 
 inputs: []
-baseCommand: /bin/false
+baseCommand: "false"
 outputs: []
 
 successCodes: [ 1 ]
--- a/v1.0/v1.0/exit-success.cwl
+++ b/v1.0/v1.0/exit-success.cwl
@@ -1,0 +1,11 @@
+#!/usr/bin/env cwl-runner
+cwlVersion: v1.0
+class: CommandLineTool
+
+inputs: []
+baseCommand: /bin/false
+outputs: []
+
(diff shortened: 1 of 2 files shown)


anton-khodak:
cwltool currently skips successCodes if the code is 2: https://github.com/common-workflow-language/cwltool/blob/52c276d1a384bed977f2355513f1c07a66b11971/cwltool/job.py#L246, needs a fix

anton-khodak:
should you add it to v1.1 as well?

mr-c:
> cwltool currently ignores successCodes if an OSError happens due to missing files: https://github.com/common-workflow-language/cwltool/blob/52c276d1a384bed977f2355513f1c07a66b11971/cwltool/job.py#L246, needs a fix

True, that should reference `errno.ENOENT` instead of `2` but I think that is correct to ignore `successCodes` here as the command was never run and there is no exit code to examine.

mr-c:
> should you add it to v1.1 as well?
We'll do another sync of v1.0 to v1.1 later

anton-khodak:
so you should change the test so that shell command is actually executed, returns a non-zero code, and successCode is actually checked

v1.0/v1.0/exit-success.cwl
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env cwl-runner
+cwlVersion: v1.0
+class: CommandLineTool
+
+inputs: []
+baseCommand: /bin/false
```
anton-khodak: can we change it to just `"false"`? it is /usr/bin/false on MacOS

mr-c: Sure

mr-c:
Thank you @anton-khodak !

PR #630, opened 2018-02-07
Using requirements instead of hints

manabuishii:
From conformance test's comment,
these files MUST use `requirements` instead of `hints`.

conformance test 8

```
tool: v1.0/cat3-tool-shortcut.cwl
doc: Test command execution in Docker with simplified syntax stdout redirection
```

conformance test 9

```
 tool: v1.0/cat3-tool-mediumcut.cwl
 doc: Test command execution in Docker with stdout redirection
```


mr-c:
Fixed in https://github.com/common-workflow-language/common-workflow-language/pull/599, thanks!

PR #638, opened 2018-02-28
Add DockerRequirement for `python` command in valueFrom-constant.cwl

anton-khodak:
Needed for REANA conformance since it uses default containers for workflows without Docker, and `frolvlad/alpine-bash` does not carry Python on board

### DIFF
--- a/v1.0/v1.0/valueFrom-constant.cwl
+++ b/v1.0/v1.0/valueFrom-constant.cwl
@@ -1,7 +1,13 @@
 class: CommandLineTool
 cwlVersion: v1.0
+
 requirements:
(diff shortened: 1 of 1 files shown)


v1.0/v1.0/valueFrom-constant.cwl
```diff
@@ -2,6 +2,8 @@ class: CommandLineTool
 cwlVersion: v1.0
 requirements:
 - class: InlineJavascriptRequirement
+- class: DockerRequirement
```
mr-c: Under `hints` please

mr-c:
Thanks!

PR #646, opened 2018-03-14
Make array-of-strings-job.yml take in an array of files

ThomasHickman:
This is to tests against errors like https://github.com/common-workflow-language/cwltool/issues/687


tetron:
Thanks!  But lets merge and release cwltool with the fix before merging the conformance test or else we're going to have a bad day.

mr-c:
Jenkins, okay to test

mr-c:
Thank you @ThomasHickman !

PR #660, opened 2018-03-23
V1.1.0 fix wf bindings

tetron:
Remove outputBinding from WorkflowOutput (it never should have been there).

Deprecate inputBInding on WorkflowInput.


tetron:
Corresponding cwltool PR https://github.com/common-workflow-language/cwltool/pull/694

v1.1.0-dev1/Process.yml
```diff
@@ -555,27 +572,25 @@ $graph:
...
-  abstract: true
   fields:
     - name: loadContents
       type:
         - "null"
         - boolean
       jsonldPredicate: "cwl:loadContents"
       doc: |
+        Use of `loadContents` in `InputBinding` is deprecated.
+        Preserved for v1.0 backwards compatability.  May be removed in
```
mr-c: "Will be removed in v2.0" and we should start that fork already (or at least a meta-issue)

tetron:
cwl-objectmodel-v1 0
cwl-objectmodel-v1 1

tetron:
I filed   #662 and #663 under the "2.0+" milestone

mr-c:
v1.1.0-dev1 is missing "Describes how to handle the inputs of a process and convert them into a concrete form for execution, such as command line parameters." for CommandLineTool.html#CommandInputParameter 's `inputBinding`

mr-c:
`CommandLineBinding` is now under `CommandInputRecordField` which doesn't make much sense. In v1.0 is is under `CommandInputParameter`

tetron:
The default behavior of schema salad is to do a depth-first traversal of the inheritance graph, sometimes this makes it do silly things.  There are options "docBefore" and "docAfter" to tweak positioning, but it might also produce better results if the traversal was breadth-first.

tetron:
Would you consider merging and fixing the ordering later?

mr-c:
I'm reviewing diffs of the HTML to detect schema oopses, so a fixed order would help. I'll see what I can do here.

tetron:
Oh, I see, let me give it a try.

tetron:
It isn't perfect but it is closer, take a look.

mr-c:
Are you also seeing the following when building the site?
```
site/cwlsite.cwl:10:9: invalid field `name`, expected one of: 'fields', 'type', 'label'
site/cwlsite.cwl:10:9: invalid field `name`, expected one of: 'fields', 'type', 'label'
site/cwlsite.cwl:24:9: invalid field `name`, expected one of: 'fields', 'type', 'label'
site/cwlsite.cwl:24:9: invalid field `name`, expected one of: 'fields', 'type', 'label'
```

tetron:
Yes, I'm seeing that too.  I think that has something to do with using an anonymous record type and it is secretly adding a "name" field to make Avro happy, but "name" is not actually valid by the spec, which is one of the things being fixed in v1.1.

mr-c:
`doc` field being an array of strings is not in the changelog

mr-c:
> 5.1.7 CommandInputRecordSchema

field | type | required | description
-- | -- | -- | --
type | enum_d9cba076fca539106791a4f46d198c7fcfbdb779 | True | Must be record

tetron:
Argh, that is an anonymous enum type for which it generates a random name.  We should probably not have any anonymous types in the spec.

mr-c:
`inputBinding` being part of `CommandInputRecordSchema` is missing from the changelog, also missing a description field; same for `CommandInputEnumSchema` (which also shows the same `enum_d062602be0b4b8fd33e69e29a841317b6ab665bc` issue)

v1.1.0-dev1/CommandLineTool.yml
```diff
@@ -51,6 +51,7 @@ $graph:
         * Default values for some fields are now expressed in the schema.
         * When defining record types, fields of type `File` may now include `secondaryFiles` and `format`.
```
mr-c: and `loadContents`, and `streamable`

mr-c:
`CommandOutputParameter.CommandOutputBinding` is has lost its description

mr-c:
Changelog is missing that `SchemaDefRequirement`  types are now `array<CommandInputRecordSchema | CommandInputEnumSchema | CommandInputArraySchema>` and not `array<InputRecordSchema | InputEnumSchema | InputArraySchema>`

mr-c:
- [ ] Need to document that `Workflow.inputs` is now `array<WorkflowInputParameter> | map<WorkflowInputParameter.id, WorkflowInputParameter.type> | map<WorkflowInputParameter.id, WorkflowInputParameter>` and not  `array<InputParameter> | map<InputParameter.id, InputParameter.type> | map<InputParameter.id, InputParameter>` 

-  [ ] Clarification about `OutputRecordField` should be present in the Workflow changelog as well
- [ ] add to Workflow changelog that `WorkflowStepInput` now has `loadContents`
- [ ] restore `WorkflowStepOutput.id`'s description to include "This is the identifier to use in the `source` field of `WorkflowStepInput` to connect the output value to downstream parameters."

mr-c:
in `Workflow`, `SchemaDefRequirement` now refers to `<CommandInputRecordSchema | CommandInputEnumSchema | CommandInputArraySchema>` which is a bit weird but maybe necessary ?

mr-c:
-[ ] `ExpressionTool.inputs` is now `array<WorkflowInputParameter> | map<WorkflowInputParameter.id, WorkflowInputParameter.type> | map<WorkflowInputParameter.id, WorkflowInputParameter>`

v1.1.0-dev1/CommandLineTool.yml
mr-c: Fixed

PR #669, opened 2018-04-17
replace uninformative syntax badge with "donate"

mr-c:
Won't render on GitHub.com but should render on commonwl.org


mr-c:
Thanks!

PR #684, opened 2018-05-02
Fewer conformance reqs

mr-c:
There is no need to require `javac` for non container using systems


mr-c:
This is ready for review and merging @tetron

PR #689, opened 2018-05-14
Add python docker to stage-unprovided-file test

Horneth:
This test seems like it should have a docker `hint` since it uses `python` as the `baseCommand` ? Otherwise it implicitly requires the machine to have python installed.


mr-c:
Thanks @Horneth !

Issue #679, opened 2018-04-24
Conformance test valueFrom with null input


tetron:
From @danbills:

> I'm a bit puzzled by the expectation of test #122, "Test that expression engine does not fail to evaluate reference to self with unprovided input"
This test has an expression $(self.basename) with an optional file input that is not provided
the test just says that it doesn't "fail" but I'm wondering how the JS engine should expect to evaluate it.
One might argue the JS engine shouldn't evaluate it at all, but that behavior breaks test #67, which has the exact same scenario except the expression is just a string, i.e. it doesn't evaluate self:
$("a ")$("string")

danbills:
This should be closed now as #694 is merged

PR #694, opened 2018-05-21
remove test that evaluates self despite no input

danbills:
as a result of #679


mr-c:
Jenkins, okay to test

PR #701, opened 2018-05-28
$5 in the baseCommand nor the arguments list should be evaluated

mr-c:


mr-c:
Summary:
- cwltool: Passes
- cwlexec: Fails (as expected)
- rabix: Passes
- Cromwell: fails due to lack of `basename` in the File objects that are returned; otherwise it is correct

PR #711, opened 2018-06-08
Add class information to tags

manabuishii:
Add tag to tags

CommandLineTool -> command_line_tool
ExpressionTool -> expression_tool
Workflow -> workflow

### DIFF
--- a/v1.0/conformance_test_v1.0.yaml
+++ b/v1.0/conformance_test_v1.0.yaml
@@ -6,7 +6,7 @@
       example_human_Illumina.pe_1.fastq,
       example_human_Illumina.pe_2.fastq]
   doc: General test of command line generation
-  tags: [ required ]
+  tags: [ required, command_line_tool ]
 
 - output:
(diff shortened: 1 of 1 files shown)


manabuishii:
If there are tags about `class` information in tags[] , it's very useful for developer.

manabuishii:
`v1.0/string-job.json` does not have `tags`
So I add just

```
tags: [ command_line_tool ]
```

https://github.com/common-workflow-language/common-workflow-language/pull/711/files#diff-d245103366eb4fae8c00d02ce3d02799R547

v1.0/conformance_test_v1.0.yaml
```diff
@@ -544,14 +544,15 @@
 
   tool: v1.0/iwdr-entry.cwl
   doc: Test if trailing newline is present in file entry in InitialWorkDir
+  tags: [ command_line_tool ]
```
mr-c: Good catch, thanks!

PR #712, opened 2018-06-10
v1.1.0-dev1 has been moved to cwl-v1.1 repository.

tetron:
Remove it to minimize confusion.  When v1.1 is relased, consider
bringing it back in via git submodule or subtree.


mr-c:
Should we update the README or add a document about how new versions are developed?

PR #731, opened 2018-07-13
Add id number to conformance test

manabuishii:
Problem:
When develop with conformance test,
developer specify number of test `-n1,2,3`.
There is no number in conformance test.
So it is hard to find specific test.

And if new conformance test is insert between existing tests,
The number is slide , brake CI.

Solution:
Each test has own number.


mr-c:
Thank you @manabuishii !

PR #737, opened 2018-07-16
Add conformance test for excess carets in a secondaryFile name

anowlcalledjosh:
Extra carets at the start of a secondaryFile should have no effect on the resulting filename.

See common-workflow-language/cwltool#824.


mr-c:
Jenkins, add to testlist

anowlcalledjosh:
I noticed that cwltool also fails to handle the case where a secondaryFile is only composed of carets, so I added another conformance test for that.

PR #738, opened 2018-07-17
Change test case doc

manabuishii:
There is no `docker` requirement.
If `Docker-based` is needed, test case must have requirement `docker`


mr-c:
Thank you @manabuishii !

PR #745, opened 2018-07-19
Add description for `self` in WorkflowStepInput when it is in `scatter`

tom-tan:
This request adds a description the value of `self` must be an element of the value of `source` field rather than the value of `source` field when the step id is specified in `scatter` field.

It makes easier to understand how `self` must behave in the case of scatter.


mr-c:
I agree, thank you. This is backed up by the conformance tests and existing engine behaviour and is implied by the following paragraph so it is good to make it more explicit.

PR #746, opened 2018-07-23
Allow passing a timeout value to cwltest.

mcovarr:
Allows passing a timeout value to override the default cwltest 10 minute timeout limit. The motivation here is to support a cloud backend where a series of steps may occasionally take longer than 10 minutes due to the vagaries of the cloud but still eventually succeed.


mr-c:
Jenkins, okay to test

mr-c:
Thank you @mcovarr !

PR #747, opened 2018-07-24
Add description about input array is empty

manabuishii:
`doc` of conformance test 127 (empty-array-input.cwl, empty-array-job.json) says below

```
doc: "Test that empty array input does not add anything to command line"
```

But there is no documentation about this in the spec.


mr-c:
Thank you @manabuishii.

I agree that this behaviour is not documented.

The test uses an empty array, which by a strict reading the of the current spec should cause the prefix to be present on the command line; but that isn't useful behavior and is not the practice of any of the CWL implementations.

(If the test used `null` instead of an empty array then the entire `inputBinding` would not have been consulted)

I think this is a fair clarification to make but I want @tetron review on this first

PR #750, opened 2018-07-27
WorkflowStepInput.valueFrom: if self is type `Directory`, then `self.…

mr-c:
…listing` should be present

### DIFF
--- a/v1.0/v1.0/single_file_from_dir.cwl
+++ b/v1.0/v1.0/single_file_from_dir.cwl
@@ -1,0 +1,28 @@
+#!/usr/bin/env cwl-runner
+cwlVersion: v1.0
+class: Workflow
+
+requirements:
+  StepInputExpressionRequirement: {}
+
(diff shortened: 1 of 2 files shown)


v1.0/v1.0/single_file_from_dir.cwl
```diff
@@ -0,0 +1,28 @@
...
+    run: dir3.cwl
+    in:
+      inf: inf
+    out: [ outdir ]
+  second:
+    run: cat-tool.cwl
+    in:
+      file1:
+        source: first/outdir
+        valueFrom: $(self.listing[0])
```
jirikuncar: Can you access the file by a name? The index might not be very reliable or desired.

stain: not as easily as one would have to filter on `basename` - but this test only needs to verify that self.listing is populated and so `[0]` should be present.

PR #762, opened 2018-08-20
replace "--reverse" with "-r" in sort

wgerlach:
The sort command under alpine /  busybox does not support the double dash argument "--reverse", it only supports "-r". Since "-r" should work in any sort implementation, I am suggesting to change the CommandLineTool sorttool.cwl/reverse accordingly.


mr-c:
Jenkis, okay to test

PR #769, opened 2018-09-07
Explicitly run `python2` for Python 2 scripts

anowlcalledjosh:
This helps avoid issues when in a Python 3 virtualenv (in which case
`python` == `python3`) and Docker isn't available.

### DIFF
--- a/v1.0/conformance_test_v1.0.yaml
+++ b/v1.0/conformance_test_v1.0.yaml
@@ -564,7 +564,7 @@
             {
                 "location": "input.txt.idx1",
                 "class": "File",
-                "checksum": "sha1$553f3a09003a9f69623f03bec13c0b078d706023",
+                "checksum": "sha1$1f6fe811644355974cdd06d9eb695d6e859f3b44",
                 "size": 1500
             },
             {
--- a/v1.0/v1.0/binding-test.cwl
+++ b/v1.0/v1.0/binding-test.cwl
@@ -28,5 +28,5 @@ inputs:
 outputs:
   args: string[]
 
-baseCommand: python2
+baseCommand: python
 arguments: ["bwa", "mem"]
--- a/v1.0/v1.0/binding-test.cwl
+++ b/v1.0/v1.0/binding-test.cwl
@@ -28,5 +28,5 @@ inputs:
 outputs:
   args: string[]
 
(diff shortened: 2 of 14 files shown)


mr-c:
Can we just make args.py py2/3 friendly instead? Not every py2 system has `python2` available (but they should)

anowlcalledjosh:
afaict, args.py is fine, it's search.py that seemed to be causing me trouble. In the end it was index.py that was iterating over a set. I *think* I've fixed it now.

v1.0/conformance_test_v1.0.yaml
```diff
@@ -564,7 +564,7 @@
             {
                 "location": "input.txt.idx1",
                 "class": "File",
-                "checksum": "sha1$553f3a09003a9f69623f03bec13c0b078d706023",
+                "checksum": "sha1$1f6fe811644355974cdd06d9eb695d6e859f3b44",
```
mr-c: Is this due to the `sorted`?

anowlcalledjosh: Yes – set iteration order in Python 2.7 appears to be consistent, but not sorted, and different to Python 3. Here's a diff between the produced index in Python 2 and Python 3 (before I put the `sorted` in):

```diff
$ colordiff -U0 working/input.txt.idx1 broken/input.txt.idx1
--- working/input.txt.idx1	2018-09-07 12:11:07.000000000 +0100
+++ broken/input.txt.idx1	2018-09-07 12:10:12.000000000 +0100
@@ -1 +1 @@
-a: 9, 3, 4, 6, 12
+a: 3, 4, 6, 9, 12
@@ -47 +47 @@
-in: 2, 14, 6, 15
+in: 2, 15, 6, 14
@@ -51 +51 @@
-is: 12, 4, 14, 6
+is: 4, 12, 6, 14
@@ -53 +53 @@
-it: 9, 11, 4, 6, 15
+it: 4, 6, 9, 11, 15
@@ -58 +58 @@
-me: 16, 1, 2, 10, 9
+me: 1, 2, 9, 10, 16
```

mr-c:
Jenkins, test this please

mr-c:
Thank you @anowlcalledjosh !

PR #787, opened 2018-10-25
SchemaDefRequirement with $import, and packed

mr-c:
Inspired by https://github.com/broadinstitute/cromwell/issues/4103 and @jeremiahsavage

### DIFF
--- a/v1.0/v1.0/touch.cwl
+++ b/v1.0/v1.0/touch.cwl
@@ -1,0 +1,21 @@
+#!/usr/bin/env cwl-runner
+cwlVersion: v1.0
+class: CommandLineTool
+
+hints:
+  DockerRequirement:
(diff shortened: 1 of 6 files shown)


v1.0/v1.0/touch.cwl
```diff
@@ -0,0 +1,21 @@
+#!/usr/bin/env cwl-runner
+cwlVersion: v1.0
+class: CommandLineTool
+
+hints:
+  DockerRequirement:
+    dockerPull: debian:stretch-slim
```
tom-tan: IMO it will be better to use `requirements` rather than `hints` because the behavior of `hints` depends on the platform and environment.

stain: But for this particular conformance test the use of Docker or not is not essential, so I would leave it as `hints`

PR #797, opened 2018-11-16
Fix scatter_embedded_subworkflow conformance test

tetron:
Looks like there was a merge conflict over count-lines13-wf.cwl and it
was resolved incorrectly.  Added count-lines18-wf.cwl which presumably
tests what the test was supposed to test.


mr-c:
Thanks @tetron

PR #812, opened 2018-12-11
Fix id for schema-def_anonymous_enum_in_array

tom-tan:
This request is to fix `id` field for the test with the label `schema-def_anonymous_enum_in_array`.
Currently its `id` is `196` but it is duplicated with the test with the label `anonymous_enum_in_array`.


tom-tan:
The conformance-testing fails but it is because of the network problems in Jenkins rather than the problem of this PR.

mr-c:
Thank you @tom-tan !

PR #824, opened 2019-01-30
Remove unconnected input

bogdang989:
Workflow input File2 is not connected anywhere.


mr-c:
As we discussed, this is not disallowed by the CWL v1.0 standards.

Issue #591, opened 2018-01-11
need "negative" test for missing `InitialWorkDirRequirement.listing[0].writable=true`


mr-c:
https://github.com/common-workflow-language/common-workflow-language/pull/589#pullrequestreview-88204906

PR #830, opened 2019-02-12
Test for InitialWorkDirRequirement.listing defaults to read-only

dleehr:
Adds a conformance test, declaring that a job should fail if a process tries to write to an InitialWorkDir that has not been marked writable.

For #591

### DIFF
--- a/v1.0/conformance_test_v1.0.yaml
+++ b/v1.0/conformance_test_v1.0.yaml
@@ -2590,3 +2590,11 @@
   doc: Test an anonymous enum inside an array inside a record, SchemaDefRequirement
   tags: [command_line_tool, schema_def]
 
+- doc: Test InitialWorkDirRequirement without writable fails on write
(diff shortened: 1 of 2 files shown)


dleehr:
Looks to be a duplicate of https://github.com/common-workflow-language/common-workflow-language/pull/621, so perhaps I'm missing something?

mr-c:
For `cwltool` this tests fails (the workflow finishes) unless `--default-container debian` is added :crying_cat_face:

mr-c:
@dleehr This PR should be made against https://github.com/common-workflow-language/cwl-v1.1

Also, we should fix `cwltool` to provide the same experience when a Docker container isn't in use

mr-c:
Which is what the other PR is blocking on as well (it dates from when v1.1.x lived in this repository)

dleehr:
OK thanks for the clarification.

PR #848, opened 2019-04-05
Add --tags option to run_test.sh

tom-tan:
This request adds `--tags` option to test only with the specified tags.
It is useful when we want to check the limited number of test cases (e.g., test only with the `required` tag).


tom-tan:
Thanks!

Issue #135, opened 2015-09-29
scattering environments as parameters


satra:
`use case:`

for many settings we often want to ensure that the same workflow runs under different environments and versions of software. one way is to build docker containers for each version (a la travis). however, in many practical cases we may want to run such regression testing on our own machines and clusters.

we have been working on an extension of `unittest` called [testkraut](https://testkraut.readthedocs.org/en/latest/) that we were considering moving over to CWL and spent some time hacking and understanding how we could.

`current hack:`

create a wrapper cwl that takes as an input an environment spec/docker image id, then the internal tool writes a cwl file with the environment or requirements based on the parameter and runs the workflow.

tetron:
It sounds like you want to parameterize the process requirements.  You can already do this for environment variables and configuration files. You do need to translate each environment into a CWL description, as CWL discourages the "run a magic bash script" approach to configuring the environment.  Does this make sense?  If you can describe a specific case of something you do now I can provide an example in CWL and see if it meets your needs.

satra:
@tetron - actually both travis and circleci specs kind of cover this idea. for example in travis we can do something like a testing matrix.

http://docs.travis-ci.com/user/customizing-the-build/#Build-Matrix

satra:
@tetron -  as a follow up to this:

> You do need to translate each environment into a CWL description, as CWL discourages the "run a magic bash script" approach to configuring the environment. 

could you please clarify this?

i can't see how you will avoid this without restricting folks to use docker images and vms. 

for example, even installing a python environment through say conda - from continuum.io - would require a sequence of statements. and since package management across tools and environments is not a solved issue, it seems that the only way out is to use images, which would have required some script (bash, fabric, ansible, dockerfile, etc.,.) to configure.

tetron:
I see, I think we're talking about slightly different things.

What I was talking about is that to configure a tool to use an existing, installed environment, you should use CWL `EnvVarRequirement` or `CreateFileRequirement` to set the specific environment variables or configuration files the tool needs.  The "magic bash script" I'm referring to is the typical environment setup script that mostly just sets environment variables (sometimes with some noisy conditional logic in an attempt to make it portable).  For example virtualenv "activate": despite being 80 lines long all it does that actually matters for running Python programs is set `VIRTUAL_ENV` and `PATH`.

[...]

PR #849, opened 2019-04-09
Set root workflow for conformance #135

aednichols:
The packed CWL in this test has two possible entry points, `main` and `touch.cwl`.

Other test cases with multiple entry points specify a root workflow, so I think I'm doing something reasonable here by making this case consistent with the others.


tetron:
Note in CWL v1.1 formalizes using "main" as the entry point if not specified.

aednichols:
Good to know, thanks

PR #860, opened 2019-06-03
remove doubled "hints"

mr-c:


v1.0/v1.0/envvar3.cwl
```diff
@@ -6,6 +6,9 @@ outputs:
   results:
     type: File
     outputBinding: { glob: results }
+  # log:
+  #   type: File
+  #   outputBinding: { glob: log }
```
tetron: Better to remove it than comment it out?

v1.0/v1.0/imported-hint.cwl
```diff
@@ -1,9 +1,6 @@
 #!/usr/bin/env cwl-runner
 cwlVersion: v1.0
 class: CommandLineTool
-hints:
```
tetron: Is this fixed in v1.1?  Looks vaguely familiar.

v1.0/v1.0/envvar3.cwl
mr-c: It is nice when one is debugging this particular test.

v1.0/v1.0/imported-hint.cwl
mr-c: Yes, https://github.com/common-workflow-language/cwl-v1.1/blob/83bb02793b59cb039bbf12f88812678e920f727e/tests/imported-hint.cwl

PR #882, opened 2019-09-16
Add `stdout` and `stderr` as symbols for CWLType

kaushik-work:
`stdout` and `stderr` were missing as symbols for CWLType


tetron:
This is wrong, the `stdout` type symbol is defined here: https://github.com/common-workflow-language/common-workflow-language/blob/main/v1.0/CommandLineTool.yml#L439

PR #889, opened 2019-11-20
Abstract operation Architecture Decision Record

mr-c:
Documents current thinking for https://github.com/common-workflow-language/common-workflow-language/issues/337

### DIFF
--- a/design-decisions/operation-2019.md
+++ b/design-decisions/operation-2019.md
@@ -1,0 +1,79 @@
+# Abstract Operations (2019)
+
+## Abstract
+
+This is a documentation of the decisions made in the design of _abstract operations_ outlined in 
+[issue #337](https://github.com/common-workflow-language/common-workflow-language/issues/337)
+For the specification of the feature please see that issue and the CWL specifications document.
+
+## Motivation
+
+**Use case 1:** While writing or modifying a CWL workflow, 
+sometimes a `CommandLineTool` or `Workflow` does not yet exist for a desired 
+step, while the rest of the workflow steps has existing tools to use.
+
+**Use case 2:** Often there are multiple implementations that can perform a desired functionality, but they may 
+have slight differences in their command line arguments or file requirements.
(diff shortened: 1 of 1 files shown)


stain:
Please help review/edit! [Readable version here](https://github.com/common-workflow-language/common-workflow-language/blob/abstract-operation/design-decisions/operation-2019.md)

stain:
Schema Salad changes for adding `Operation` in common-workflow-language/cwl-v1.2#3 and its implementation in common-workflow-language/cwltool#1218

kaushik-work:
@stain 



## Argument A

Why can this not be achieved by having place holder regular process objects that have and empty list for steps or empty list for base command? I feel this is about the same amount of typing with the advantage that then I can go in and just flesh out the skeleton. This is like when we create program skeletons and do things like 

```
def foo(in1, in2):
  return True
```

and come back later and fill out the stub.


The use case for having multiple tools with the same interface is intriguing but I feel that this is not a compelling use case for me. Related use cases are elegantly handled by the existing typedef system.


## Argument B

What is the behavior of the executor when it encounters this abstract operation. Does it fill out the outputs with nulls? With some random value of the correct type?

Thanks!

tetron:
> The use case for having multiple tools with the same interface is intriguing but I feel that this is not a compelling use case for me.

However, that's the use case that motivated the original proposal.  @stain's use cases focus more on the placeholder aspect, though.  If it is a no-op CommandLineTool you can't programmatically determine if something is a placeholder unless we documented a specific constraints on the CommandLineTool type that made it a placeholder --- but then we might as well define a new record type anyway so it can be typechecked.

> What is the behavior of the executor when it encounters this abstract operation. Does it fill out the outputs with nulls?

Perhaps the upfront static checker should declare it invalid to run.  Or attempting to run the step should fail.  Or, it could be treated like a skipped step and produce nulls.  It's a good question.

PR #894, opened 2020-01-21
Fix tag for StepInputExpressionRequirement

tom-tan:
Most tests with `StepInputExpressionRequirement` uses `step_input` tag but the test 111 uses `step_input_expression`.
This request fixes it.


mr-c:
Thanks!

PR #902, opened 2020-06-02
Tests: bugfix: File class objects should use `location`, not `path`.

cure:
This fixes test 150, 151 and 184 on Arvados.


mr-c:
Hmm.. We don't really make changes to the v1.0.x series anymore. Arvados used to pass these tests, what happened?

PR #916, opened 2020-11-13
Any does not include nulls

mr-c:
There is a bug in cwltool; it allows an `ExpressionTool` to return `null` for an `Any`, but rejects the same from `cwl.output.json` output reporting from a `CommandLineTool`.

@tetron How shall we handle this? Fix all occurrences of `Any` that should be `Any?` in the v1.0 conformances tests and release CWL v1.0.3 ?


mr-c:
Corresponding cwltool bug report https://github.com/common-workflow-language/cwltool/issues/1369

PR #929, opened 2021-09-19
Clarify $() and ${}

### LINKED ISSUE #206: clarify $(..) vs ${...}
It isn't obvious when to use one or the other in the current spec: http://www.commonwl.org/draft-3/CommandLineTool.html#Expressions

kinow:
Closes #206 

Hi,

Having a go at the Barn Raising issue about ES expression and ES function body. At first I wrote a longer version with an example `$(new Date())` for expression, and a longer `${var dt = new Date(); dt.setDate(dt.getDate() - 2); return dt}` function body that returns a date minus 2 days.

But I think it became too specific and made the section even longer? I also thought that perhaps the comparison could go into a NOTE section, but didn't find any such section nearby, so went with the simpler approach.

Happy to amend as per feedback :+1: 

Thanks
Bruno

p.s.: this PR also fixes a typo, "EMACScript" where it should be "ECMAScript".

### DIFF
--- a/draft-3/concepts.md
+++ b/draft-3/concepts.md
@@ -307,16 +307,20 @@ requirement `InlineJavascriptRequirement`.  Expressions may be used in any
 field permitting the pseudo-type `Expression`, as specified by this
 document.
 
-Expressions are denoted by the syntax `$(...)` or `${...}`.  A code
-fragment wrapped in the `$(...)` syntax must be evaluated as a
-[ECMAScript expression](http://www.ecma-international.org/ecma-262/5.1/#sec-11).  A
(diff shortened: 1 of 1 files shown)


mr-c:
Thank you very much @kinow !

We aren't making changes to the older versions of CWL anymore. Can you make a new pull request in the https://github.com/common-workflow-language/cwl-v1.2/ repository against the https://github.com/common-workflow-language/cwl-v1.2/tree/1.2.1_proposed branch?

kinow:
Done @mr-c .

Closing as superseded by https://github.com/common-workflow-language/cwl-v1.2/pull/123 I think

PR #946, opened 2022-07-07
add CACM paper

mr-c:


CITATION
```diff
@@ -1,4 +1,27 @@
-To cite the Common Workflow Language standard in a publication, please use:
+To cite the Common Workflow Language  and the CWL project in a scholary work, please use:
```
kinow, line 1: Extra space but I believe this is fixed by Jekyll/markdown parsers/etc (`Language  and`)

kinow (approved):
Read it twice searching for misspellings, formatting issues, checking author names. Everything looks OK. There's some extra space but I believe that gets fixed by markdown parsing tools. +1

PR #960, opened 2024-08-29
Clarify saladVersion

tom-tan:
This request clarifies `saladVersion`.
It must be `v1.3` because the schema includes union and map schemas.


tom-tan:
Oh, sorry.
I found that the `main` branch does not use map and union schemas but `codegen` branch does.
Can I leave this request as is to clarify the schema in main branch is SALAD v1.3 or should I re-send another request for `codegen` branch rather than the `main` branch?

PR #962, opened 2025-12-20
[codegen] ProcessGenerator extension: match parameter specialization to ExpressionTool

mr-c:
Otherwise the outputs will have no `type` field


GlassOfWhiskey (approved):
LGTM

